PR #32, opened 2016-10-19
App Status

austek-mdsol:
Add App Status and build info to it

@crudd-mdsol @mdrynkorn-mdsol @eyomi-mdsol @rmillward-mdsol


crudd-mdsol:
LGTM

PR #40, opened 2017-10-03
This reduces the size of the build for starting the proxy and gets ri…

johngluckmdsol:
We're starting to use this proxy for making mauth calls using the client.  This makes it easier and faster


jprice-mdsol:
I'm happy with this - just waiting for Ali to have a look tomorrow (he's in the UK)

PR #44, opened 2017-10-17
Adding gitattributes file to set docker script to eol=lf.

fbotero-mdsol:
Windows uses `crlf` which causes issues with the bash magic number.


mnohai-mdsol:
LGTM

PR #47, opened 2017-10-20
[MCC-344550] Move to SBT

austek:
Move project to sbt to enable creation of clients for scala


austek:
I've tried to make maven work with scala but scalatest and some scala futures didn't work. So we don't have much choice TBH

johngluckmdsol:
I just realized that we should probably change the docker base image here to use something that has sbt installed

mnohai-mdsol:
This is a standalone app, what kind of clients are you trying to build off of it?

austek:
By Clients I mean to have akka-http implementations of authenticator and signer

johngluckmdsol:
@austek, right now we are building the project and then running it but it may make sense to pull the jar from Artifactory.  But yeah, that sounds like it could work.  @benton posted some instructions on the docker-mentor slack channel for how to get our images into AWS.  Perhaps that is the way to go

PR #64, opened 2018-07-20
[MCC-403926] Replace getForKeyCaseInsensitive for getForKey

mcallonalvarez:
Fix an issue when using the implicit method "fromSignedRequestToHttpRequest" which is ignoring Https methods other than GET because the method akka.http.scaladsl.model.HttpMethods.getForKeyCaseInsensitive("POST") is returning None. 

This method has been replaced for akka.http.scaladsl.model.HttpMethods.getForKey("POST") which returns the "HttpMethod(POST)" value correctly.

### DIFF
--- a/modules/mauth-signer-akka-http/src/test/scala/com/mdsol/mauth/http/ImplicitsTest.scala
+++ b/modules/mauth-signer-akka-http/src/test/scala/com/mdsol/mauth/http/ImplicitsTest.scala
@@ -1,0 +1,34 @@
+package com.mdsol.mauth.http
+
+import java.net.URI
+
+import com.mdsol.mauth.http.Implicits._
+import com.mdsol.mauth.{SignedRequest, UnsignedRequest}
+import org.scalatest.{Matchers, WordSpec}
+
+class ImplicitsTest extends WordSpec with Matchers {
+
+  "Implicits fromSignedRequestToHttpRequest" should {
+
+    "Generate a POST HttpRequest from a SignedRequest" in {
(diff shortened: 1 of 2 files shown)


modules/mauth-signer-akka-http/src/test/scala/com/mdsol/mauth/http/ImplicitsTest.scala
```diff
@@ -0,0 +1,34 @@
...
+class ImplicitsTest extends WordSpec with Matchers {
+
+  "Implicits fromSignedRequestToHttpRequest" should {
+
+    "Generate a POST HttpRequest from a SignedRequest" in {
+      val signedRequest = SignedRequest(UnsignedRequest(httpMethod = "POST", uri = new URI("/")), "X_MWS_AUTHENTICATION_HEADER_VALUE", "X_MWS_TIME_HEADER_VALUE")
+      fromSignedRequestToHttpRequest(signedRequest).method.toString() should be("HttpMethod(POST)")
+    }
+
+    "Generate a GET HttpRequest from a SignedRequest when non-uppercase method specified" in {
```
mcallonalvarez: Is the definition of this test correct?

red-benabas:
Looks good to me
@rlichszteld-mdsol you will be needing this when updating the study environment status in Plinth from StudyManagement API

rlichszteld-mdsol:
LGTM, considering HttpMethod should be case-sensitive (https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html) case mismatch should default to "GET"

PR #66, opened 2018-07-26
[MCC-403926] Enforce always content type application/json

mcallonalvarez:
Fix POST/PATCH so instead of sending always Content Type "plain/text", enforce that Content-type  been "application/json".


red-benabas:
Looks good to me

PR #68, opened 2018-08-16
MCC-410498 encoding changed from UTF-8 to ISO-8859-1

pcsefalvay:
Changed encoding from UTF-8 to ISO-8859-1 to support upload of binary files and not just JSON strings in the request.

### DIFF
--- a/modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
+++ b/modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
@@ -70,11 +70,11 @@ public class RequestAuthenticator implements Authenticator {
     String unencryptedRequestString =
         MAuthSignatureHelper.generateUnencryptedSignature(
             mAuthRequest.getAppUUID(), mAuthRequest.getHttpMethod(), mAuthRequest.getResourcePath(),
-            new String(mAuthRequest.getMessagePayload(), StandardCharsets.UTF_8),
+            new String(mAuthRequest.getMessagePayload(), StandardCharsets.ISO_8859_1),
             String.valueOf(mAuthRequest.getRequestTime())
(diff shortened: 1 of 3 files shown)


tmilner:
I'm confused, why would limiting the character set to ISO-8859-1 (Which appears to just be a subset of the first 256 UTF-8 characters) allow you to upload binary files? Is this because checkmate doesn't support UTF-8?

prascuna:
is the server going to change its charset too?

uahmed:
As per https://tools.ietf.org/html/rfc7158#page-9

> 8.  String and Character Issues
8.1.  Character Encoding
   JSON text __SHALL be encoded in UTF-8, UTF-16, or UTF-32__.  The default
   encoding is UTF-8, and JSON texts that are encoded in UTF-8 are
   interoperable in the sense that they will be read successfully by the
   maximum number of implementations; there are many implementations
   that cannot successfully read texts in other encodings (such as
   UTF-16 and UTF-32).

In an earlier version of https://tools.ietf.org/html/rfc2616 there was a conflicting guideline
> 3.7.1 Canonicalization and Text Defaults
...... The "charset" parameter is used with some media types to define the
   character set (section 3.4) of the data. When no explicit charset
   parameter is provided by the sender, media subtypes of the "text"
   type are defined to have __a default charset value of "ISO-8859-1"__ when
   received via HTTP.

However a) this applied to only text media types and not application/json, and b) it was later removed in the next revision of this RFC https://tools.ietf.org/html/rfc7231

So in summary this change is not compatible with the standards as defined in RFC-7231. On the other hand I would expect callers to be always sending both content type and character set when making a request and server to always respect it.

jatcwang:
Closing due to this change contradicting the RFC as well as being (I believe) breaking for existing servers. Seems like the linked tickets are closed too.

PR #71, opened 2018-08-29
[MCC-414889] Fix - Add support to customize Content-Type

mcallonalvarez:
#### What is this PR for?

The following fixes have been carried out in Scala asynchronous Mauth client:
- Modify HTTP method to make it case-insensitive.
- Add support to customize content-type (apply "text/plain(UTF-8)" by default)

#### What are the relevant JIRA tickets?
[MCC-414889](https://jira.mdsol.com/browse/MCC-414889)

### DIFF
--- a/modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/MAuthRequestSigner.scala
+++ b/modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/MAuthRequestSigner.scala
@@ -15,7 +15,7 @@ import scala.util.{Failure, Success, Try}
   * @param uri        The URI of the API call , (host name and port not included)
   * @param body       The body of the request in string form
   */
-case class UnsignedRequest(httpMethod: String = "GET", uri: URI, body: Option[String] = None, headers: Map[String, String] = Map.empty)
+case class UnsignedRequest(httpMethod: String = "GET", uri: URI, body: Option[String] = None, headers: Map[String, String] = Map.empty, contentType: Option[String] = None)
 
 /**
   * Library agnostic representation of a signed request, including header data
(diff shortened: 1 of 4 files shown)


modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/MAuthRequestSigner.scala
```diff
@@ -15,7 +15,7 @@ import scala.util.{Failure, Success, Try}
   * @param uri        The URI of the API call , (host name and port not included)
   * @param body       The body of the request in string form
   */
-case class UnsignedRequest(httpMethod: String = "GET", uri: URI, body: Option[String] = None, headers: Map[String, String] = Map.empty)
+case class UnsignedRequest(httpMethod: String = "GET", uri: URI, body: Option[String] = None, headers: Map[String, String] = Map.empty, contentType: Option[String] = None)
```
austek: `Content-Type` is another header, do we really need a parameter for it?

ppoltorak-mdsol: How about passing it in a Map and later doing the same logic with difference that it will be taken from the map and set on the entity?

mcallonalvarez: Changed! Now Content-type is passed it in the headers map and removed later before the actual Akka Http call.

rlichszteld-mdsol: Cool trick! 👍

PR #89, opened 2019-10-08
[MCC-543935][MCC-543947] Added mauth v2 support for java signer & Authentication

bgong-mdsol:
This PR is to support MAuth client for MWSV2 signature for Java, it includes
1. Added necessary MWSV2 signature flow.
2. Updated common modules in Java
3. Updated mauth-signer-apache (Java)
4. Unit tests for the added V2 Signature class methods.
5. Updated Authentication & unit test (Java)
6. Updated proxy  & unit test

Notes: Authentication(Scala) and  mauth-signer-akka(Scala) will be done in follow-up PR. 

@mdsol/architecture-enablement @austek


modules/mauth-signer/src/main/java/com/mdsol/mauth/DefaultSigner.java
```diff
@@ -80,4 +114,45 @@ public DefaultSigner(UUID appUUID, PrivateKey privateKey, EpochTimeProvider epoc
 
     return headers;
   }
+
+  @Override
+  public Map<String, String> generateRequestHeadersV2(String httpVerb,
+      String requestPath, String requestPayload, String queryParameters) throws MAuthSigningException {
+
+    if (disableV1 == false && mauthVersion.equalsIgnoreCase(MAuthVersion.MWS.getValue())) {
+      return generateRequestHeaders(httpVerb, requestPath, requestPayload);
```
ykitamura-mdsol: quick question!
when `disableV1` is false, do you have both V1 and V2 headers in the request headers?

https://github.com/mdsol/mauth-client-ruby/blob/v5.0.2/spec/client/signer_spec.rb#L20-L23

bgong-mdsol: If disableV1 is false and the specified version is MWS(V1), only has V1 headers. Otherwise, will generate V2 headers only.

ykitamura-mdsol: During the transition phase, we need to support the both versions so I think you need to modify the code to have both headers when disableV1 is false.

modules/mauth-signer/src/main/java/com/mdsol/mauth/DefaultSigner.java
```diff
@@ -80,4 +114,45 @@ public DefaultSigner(UUID appUUID, PrivateKey privateKey, EpochTimeProvider epoc
...
+
+    // mAuth uses an epoch time measured in seconds
+    long currentTime = epochTimeProvider.inSeconds();
+
+    String stringTosign = MAuthSignatureHelper.generateStringToSign(
+        appUUID, httpVerb, requestPath, queryParameters, requestPayload, String.valueOf(currentTime), mauthVersion);
+
+    String encryptedSignature;
+    try {
+      encryptedSignature = MAuthSignatureHelper.encryptSignature(privateKey, stringTosign);
```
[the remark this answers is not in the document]
bgong-mdsol: Fixed in https://github.com/mdsol/mauth-jvm-clients/pull/89/commits/51b34b1f9bbc72ec164f516b827fa41eeafd9852

modules/mauth-signer/src/main/java/com/mdsol/mauth/DefaultSigner.java
```diff
@@ -80,4 +114,45 @@ public DefaultSigner(UUID appUUID, PrivateKey privateKey, EpochTimeProvider epoc
...
+    } catch (IOException | CryptoException e) {
+      logger.error("Error generating request headers", e);
+      throw new MAuthSigningException(e);
+    }
+
+    HashMap<String, String> headers = new HashMap<>();
+    headers.put(
+        MAuthRequest.MCC_AUTHENTICATION_HEADER_NAME,
+        MAuthHeadersHelper.createAuthenticationHeaderValue(appUUID, encryptedSignature, mauthVersion)
```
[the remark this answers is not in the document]
bgong-mdsol: Fixed in https://github.com/mdsol/mauth-jvm-clients/pull/89/commits/51b34b1f9bbc72ec164f516b827fa41eeafd9852

modules/mauth-signer/src/main/java/com/mdsol/mauth/DefaultSigner.java
bgong-mdsol: Fixed in https://github.com/mdsol/mauth-jvm-clients/pull/89/commits/51b34b1f9bbc72ec164f516b827fa41eeafd9852

modules/mauth-common/src/main/java/com/mdsol/mauth/util/MAuthHeadersHelper.java
```diff
@@ -1,28 +1,48 @@
 package com.mdsol.mauth.util;
 
+import com.mdsol.mauth.MAuthVersion;
+
 import java.util.UUID;
 
 public class MAuthHeadersHelper {
 
+  public static final String AUTH_HEADER_DELIMITER = ";";
```
ykitamura-mdsol: This is not a request to change but I just want to explain the reason behind adding `AUTH_HEADER_DELIMITER` in V2, as far as I understand.

Per [MAuth Protocol V2 Specification](https://docs.google.com/document/d/1X3X5V8LnNhKAqS53SfFJJnZe-TrjdMPw09tnO3ZpfdU/edit?usp=sharing):

> A semi-colon (";") to indicate the end of the digital signature (since semi-colons are not present in base64-encoding).  This is present so that in the future, the next version of the protocol could be specified on the same header after this semi-colon, e.g. MWSV2 <app_uuid>:<digital_signature_for_request>; MWSV3 <the_mwsv3_signature>;

I have no idea when we will have `MWSV3`, but if so, the signature will be added to the same `MCC-Authentication` header.

That is why I made a slightly different regex for V2 in Python (Removed `\A` and `\Z`, check for the token and `AUTH_HEADER_DELIMITER`):
https://github.com/mdsol/mauth-client-python/blob/0.2.1/mauth_client/consts.py#L13

But again, you can just kick the can 😄

ykitamura-mdsol:
If possible, can you add logging of the mauth version in the [authenticate method](https://github.com/mdsol/mauth-jvm-clients/pull/89/files#diff-e590b066f94b1297d9708e472275cb0aR57)?

In [MAuth:  Signing with Query Params](https://docs.google.com/document/d/1Ydv8fwvpvcGD8lzSo7W-8J3cTcZHQtGyZD9zR9RQ3Ac/edit?usp=sharing) doc under the "logging" section, it says:

> The version of the mauth signature validated on the server should be included in the mauth-client middleware logging so that servers log what protocol version of the client is being used.

Here is the log message in ruby:
https://github.com/mdsol/mauth-client-ruby/blob/v5.0.2/lib/mauth/client/authenticator_base.rb#L50-L54
```ruby
        logger.info(
          "Mauth-client attempting to authenticate request from app with mauth" \
          " app uuid #{object_app_uuid} to app with mauth app uuid #{client_app_uuid}" \
          " using version #{object_token}."
        )
```

modules/mauth-authenticator-apachehttp/README.md
```diff
@@ -6,7 +6,7 @@ This is an implementation of Medidata Authentication Client Authenticator to val
 
 1. Configuration
   * MAuth uses [Typesafe Config](https://github.com/typesafehub/config).
-  Create `application.conf` on your classpath with following content.
+  Create `application.conf` on your classpath with following content. The disable_v1 flag can be set to authenticate incoming requests with Mauth protocol V2 only, default to false.
```
austek: if default for `disable_v1` is `false` then it should really be part of `reference.conf`

modules/mauth-authenticator-apachehttp/README.md
```diff
@@ -49,14 +51,13 @@ This is an implementation of Medidata Authentication Client Authenticator to val
         @Override
```
austek: Instead of writing a sample code here, could we create an example project that user can see possibly another module that's not published

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/AuthenticatorConfiguration.java
```diff
@@ -8,29 +8,37 @@
...
+  private final boolean disableV1;
 
   public AuthenticatorConfiguration(Config config) {
     this(
         config.getString(BASE_URL_PATH),
         config.getString(REQUEST_URL_PATH),
         config.getString(TOKEN_URL_PATH),
-        config.getLong(TIME_TO_LIVE_SECONDS)
+        config.getLong(TIME_TO_LIVE_SECONDS),
+        config.hasPath(DISABLE_MAUTH_V1) || config.isEmpty() ? config.getBoolean(DISABLE_MAUTH_V1) : false
```
austek: just add `disable_v1` to reference.conf instead of this logic and just use `config.getBoolean(DISABLE_MAUTH_V1)`

modules/mauth-authenticator-apachehttp/README.md
austek: Also, `disable_v1` is vague and using boolean type is limiting as well. Instead, I suggest to use `enum` and more verbose values, 

May be something like:
```
enum AuthType {
    V1, V1_2, V2
}
```

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
```diff
@@ -22,15 +27,27 @@
   private final String httpMethod;
   private final long requestTime;
   private final String resourcePath;
+  private final String queryParameters;
+  private final boolean diableV1;
```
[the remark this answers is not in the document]
austek: Going back to point about boolean state of disable V1

```
enum MAuthType {
    V1, // support current clients, as they only know about V1
   V1_2, // support transition period, provide backward compantibility
   V2 //support only the next version
}
```
am I correct in my assumptions?

modules/mauth-common/src/main/java/com/mdsol/mauth/util/MAuthSignatureHelper.java
```diff
@@ -13,25 +15,101 @@
...
+   * @param resourceUrl: resource_url_path (no host, port or query string; first "/" is included)
+   * @param queryParameters: request parameters string
+   * @param requestBody: request body string
+   * @param epochTime: current seconds since Epoch
+   * @param mauthVersion: Mauth version (MSW or MSWV2)
+   * @return String
+   *
+   * @throws MAuthSigningException
+   */
+  public static String generateStringToSign(UUID appUUID, String httpMethod, String resourceUrl,
```
austek: If V2 is going to be the default I would use name `generateStringToSign` for V2 and rename this one to be `generateStringToSignV1AndV2` and possibly deprecate it so people will use it with caution

modules/mauth-signer/src/main/java/com/mdsol/mauth/Signer.java
```diff
@@ -18,6 +18,23 @@
    * @return MAuth headers which should be appended to the request before sending.
    * @throws MAuthSigningException when request cannot be signed
    */
-  Map<String, String> generateRequestHeaders(String httpVerb, String requestPath,
-                                             String requestPayload) throws MAuthSigningException;
+  Map<String, String> generateRequestHeaders(String httpVerb,
```
austek: Again, I would use `generateRequestHeaders` for the desired default of v2 and rename this and deprecate

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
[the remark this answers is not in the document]
bgong-mdsol: In the Design Decision document, it says:
by default clients sign requests with both x-mws-authentication and mcc-authentication  
…
When clients are confident that all of the servers they contact support V2, they can configure their mauth-client to only sign requests with mcc-authentication 
For signer, if the disable_v1 flag is set to false, clients will send both V1 and V2 headers.  If true, clients will send V2 headers only.
For authentication, clients will authenticate with only the highest protocol version present. if the "disable_v1 flag is set to true, clients will authenticate for V2 only.

modules/mauth-authenticator-apachehttp/README.md
bgong-mdsol: Design Decision document says:
by default clients sign requests with both x-mws-authentication and mcc-authentication
…
When clients are confident that all of the servers they contact support V2, they can configure their mauth-client to only sign requests with mcc-authentication
For signer, if the disable_v1 flag is set to false, clients will send both V1 and V2 headers. If true, clients will send V2 headers only.
For authentication, clients will authenticate with only the highest protocol version present. if the "disable_v1 flag is set to true, clients will authenticate for V2 only.

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/AuthenticatorConfiguration.java
bgong-mdsol: Updated as suggested https://github.com/mdsol/mauth-jvm-clients/pull/89/commits/c69d4044b4570262dc44715a11106302c134ac6f

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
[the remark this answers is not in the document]
bgong-mdsol: > typo 'diableV1'

fixed https://github.com/mdsol/mauth-jvm-clients/pull/89/commits/c69d4044b4570262dc44715a11106302c134ac6f

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
```diff
@@ -22,15 +27,27 @@
...
+  private final String mauthVersion;
```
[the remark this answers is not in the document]
bgong-mdsol: Updated to use MAuthVersion enum https://github.com/mdsol/mauth-jvm-clients/pull/89/commits/c69d4044b4570262dc44715a11106302c134ac6f

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
```diff
@@ -85,13 +113,22 @@ private void validateRequestTime(long requestTime) {
     }
   }
 
+  private void validateMauthVersion(boolean diableV1, String mauthVersion) {
```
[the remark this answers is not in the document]
bgong-mdsol: we do validate first, then set this.diableV1

modules/mauth-common/src/main/java/com/mdsol/mauth/util/MAuthSignatureHelper.java
bgong-mdsol: I am thinking to use generateStringToSign as a generic method to return the string to sign based on the version specified and have updated code to call **generateStringToSign** instead of generateStringToSignV2.

modules/mauth-signer/src/main/java/com/mdsol/mauth/Signer.java
bgong-mdsol: It may cause some issue with the existing modules which is calling **generateRequestHeaders** if rename it.  We need to think a better name for generateRequestHeadersV2, how about generateRequestHeadersEx?

bgong-mdsol: generateRequestHeaders should be th most suitable name, renamed the both method name as suggested. See  https://github.com/mdsol/mauth-jvm-clients/pull/89/commits/24c95e0625d79457b80781f3972c4f7f2f1e9021

CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+# 1.0.16
```
austek: please change to 
```
# Changelog
All notable changes to this project will be documented in this file.

The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),

## [Unreleased]
### Added
- Add support for MWSV2 protocol in Java modules

### Changed
- Renamed com.mdsol.mauth.Signer.generateRequestHeaders to com.mdsol.mauth.Signer.generateRequestHeadersV1
```
and possibly
```
### Deprecated
- //please add headers and V1 support here
```

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
```diff
@@ -22,15 +27,27 @@
   private final String httpMethod;
   private final long requestTime;
   private final String resourcePath;
+  private final String queryParameters;
+  private final boolean disableV1;
```
austek: Field `disableV1` seems to be assigned but never accessed, is it really needed?

bgong-mdsol: Please see refactor https://github.com/mdsol/mauth-jvm-clients/pull/89/commits/d24aad04e50a64dc504d8035d8a1a56901a81c5e

CHANGELOG.md
bgong-mdsol: Updated https://github.com/mdsol/mauth-jvm-clients/pull/89/commits/d24aad04e50a64dc504d8035d8a1a56901a81c5e

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
```diff
@@ -13,7 +15,10 @@
 
   public static final String X_MWS_TIME_HEADER_NAME = "x-mws-time";
```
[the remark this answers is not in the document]
bgong-mdsol: updated https://github.com/mdsol/mauth-jvm-clients/pull/89/commits/39b3381c53ba918c45a4bcb14bdaef64a5aed453

modules/mauth-authenticator-apachehttp/README.md
```diff
@@ -15,6 +15,7 @@ This is an implementation of Medidata Authentication Client Authenticator to val
                 
         mauth {
             base_url: "http://localhost"
+            disable_v1: false
```
austek: as this is in `reference.conf`, it doesn't have to be provided, unless default value is overridden

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
```diff
@@ -37,6 +40,18 @@ public RequestAuthenticator(ClientPublicKeyProvider clientPublicKeyProvider) {
     this(clientPublicKeyProvider, 10L);
   }
 
+  /**
+   * Uses
+   * 10L as default value for request validation timeout
+   * {@link com.mdsol.mauth.utils.ClientPublicKeyProvider} as the EpochTimeProvider
+   *
+   * @param clientPublicKeyProvider PublicKey provider
+   * @param disableV1
```
austek: please provide a description for `disableV1` param

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
```diff
@@ -47,21 +62,66 @@ public RequestAuthenticator(ClientPublicKeyProvider clientPublicKeyProvider, lon
     this(clientPublicKeyProvider, requestValidationTimeoutSeconds, new CurrentEpochTimeProvider());
   }
 
-  public RequestAuthenticator(ClientPublicKeyProvider clientPublicKeyProvider, long requestValidationTimeoutSeconds, EpochTimeProvider epochTimeProvider) {
+  /**
+   * Uses {@link com.mdsol.mauth.utils.ClientPublicKeyProvider} as the EpochTimeProvider
+   *
+   * @param clientPublicKeyProvider  PublicKey provider
+   * @param requestValidationTimeoutSeconds timeout
+   * @param disableV1
```
austek: please provide a description for `disableV1` param

bgong-mdsol: fixed https://github.com/mdsol/mauth-jvm-clients/pull/89/commits/639fbc9ed0fbf79345c5ddad83d1a8b379b68852

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
bgong-mdsol: fixed https://github.com/mdsol/mauth-jvm-clients/pull/89/commits/639fbc9ed0fbf79345c5ddad83d1a8b379b68852

modules/mauth-authenticator-apachehttp/README.md
bgong-mdsol: fixed https://github.com/mdsol/mauth-jvm-clients/pull/89/commits/639fbc9ed0fbf79345c5ddad83d1a8b379b68852

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
bgong-mdsol: Fixed

PR #92, opened 2019-11-02
[MCC-543936][MCC-561308] mauth v2 support scala

bgong-mdsol:
This PR is to support MAuth client scala modules for mauth v2, it includes

Updated mauth-authentication(Scala) & unit test;
Updated mauth-signer-akka(Scala) & unit test;
Updated mauth-authenticator-akka(Scala) & unit test

Related DOCs:
https://learn.mdsol.com/display/CA/MAuth%20Protocol%20V2%20Specification
https://mdsol.jiveon.com/community/research-development/rd-guilds/architecture-guild/architecture-enablement-%C3%A6/blog/2019/10/14/mauth-checkpoints-for-adding-protocol-v2-support-in-clients#jive_content_id_Signing_and_verifying_algorithm_is_now_using_the_standard_RSA_methods 

@austek 
@mdsol/architecture-enablement 
@dgeyer-mdsol 
@marrodriguez


modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
```diff
@@ -112,6 +167,32 @@ trait MAuthDirectives extends StrictLogging {
       None
     }
   }
+
+  private def getQueryString(req: HttpRequest): String =
+    if (req.uri.rawQueryString == None) "" else req.uri.rawQueryString.get
+
+  private def validateAuthenticationValue(headerName: String, authentication: String): Directive1[String] = {
+    val mauthVersion: MAuthVersion = if (headerName eq `MCC-Authentication`.name) MAuthVersion.MWSV2 else MAuthVersion.MWS
```
austek: This doesn't feel right, `headerName` is a string and can be anything really, it's not safe currently

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
```diff
@@ -184,4 +197,258 @@ class MAuthDirectivesSpec extends WordSpec with Matchers with ScalatestRouteTest
       }
     }
   }
+
+  "authenticate for V2" should {
+    //lazy val route: Route = authenticate(executor, authenticatorV2, timeout, requestValidationTimeout).apply(complete(HttpResponse()))
```
austek: no comments

austek:
Could you also fix following line please
https://github.com/mdsol/mauth-jvm-clients/blob/develop/modules/mauth-proxy/src/main/java/com/mdsol/mauth/proxy/ProxyConfig.java#L25

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
```diff
@@ -92,9 +100,56 @@ trait MAuthDirectives extends StrictLogging {
   //  Utility functions
   /////////////////////////////////////////////
 
-  private def extractAuthHeaderDetail(str: String): Option[AuthHeaderDetail] = {
-    if (str.startsWith("MWS ")) {
-      str.replaceFirst("MWS ", "").split(":").toList match {
+  def extractLatestAuthenticationHeader(disableV1: Boolean): Directive1[String] = {
```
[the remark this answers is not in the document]
bgong-mdsol: updated in https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/bfa021f78a935387f61abda605b5bddcedd1f62a

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
```diff
@@ -92,9 +100,56 @@ trait MAuthDirectives extends StrictLogging {
...
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+
+  def extractLatestTimeHeaderDirective(disableV1: Boolean): Directive1[Long] = {
```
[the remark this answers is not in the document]
bgong-mdsol: updated in https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/bfa021f78a935387f61abda605b5bddcedd1f62a

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
```diff
@@ -92,9 +100,56 @@ trait MAuthDirectives extends StrictLogging {
...
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+
+  private def extractAuthHeaderDetail(str: String, mauthVersion: MAuthVersion): Option[AuthHeaderDetail] = {
+    val regexString = mauthVersion.getValue + " "
```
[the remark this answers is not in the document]
bgong-mdsol: updated https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/bfa021f78a935387f61abda605b5bddcedd1f62a

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
```diff
@@ -112,6 +167,32 @@ trait MAuthDirectives extends StrictLogging {
       None
     }
   }
+
+  private def getQueryString(req: HttpRequest): String =
+    if (req.uri.rawQueryString == None) "" else req.uri.rawQueryString.get
```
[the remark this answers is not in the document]
bgong-mdsol: updated in https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/bfa021f78a935387f61abda605b5bddcedd1f62a

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
bgong-mdsol: check if the headerName is MCC-Authentication or X-MWS-Authentication first, see the fix in  https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/bfa021f78a935387f61abda605b5bddcedd1f62a

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
```diff
@@ -67,7 +75,7 @@ trait MAuthDirectives extends StrictLogging {
 
   val extractMAuthHeader: Directive1[AuthHeaderDetail] =
```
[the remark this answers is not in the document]
bgong-mdsol: debricated the both extractMwsTimeHeader and extractMwsAuthenticationHeader in  https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/0372d76b0ad8eacbd330b8fe357918f8c8d9dd11

It seems that extractMAuthHeader is in used by some modules, added  extractMCCAuthenticationDetail for V2.

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
```diff
@@ -92,9 +100,56 @@ trait MAuthDirectives extends StrictLogging {
...
-  private def extractAuthHeaderDetail(str: String): Option[AuthHeaderDetail] = {
-    if (str.startsWith("MWS ")) {
-      str.replaceFirst("MWS ", "").split(":").toList match {
+  def extractLatestAuthenticationHeader(disableV1: Boolean): Directive1[String] = {
+    optionalHeaderValueByType[`MCC-Authentication`]((): Unit).flatMap { hdr =>
+      Try(hdr.get.value).toOption match {
+        case Some(auth: String) =>
+          validateAuthenticationValue(`MCC-Authentication`.name, auth)
+        case None => {
+          if (disableV1) {
```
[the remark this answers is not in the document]
bgong-mdsol: Fixed in https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/0372d76b0ad8eacbd330b8fe357918f8c8d9dd11

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
```diff
@@ -139,7 +178,205 @@ class MAuthDirectivesSpec extends WordSpec with Matchers with ScalatestRouteTest
...
+    "pass successfully authenticated request with parameters" in {
+      (client.getPublicKey _).expects(appUuid).returns(Future(Some(publicKey)))
+      //noinspection ConvertibleToMethodValue
+      (mockEpochTimeProvider.inSeconds _: () => Long).expects().returns(timeHeader)
+
+      Get("/").withHeaders(
+        RawHeader(MAuthRequest.MCC_TIME_HEADER_NAME, timeHeader.toString),
+        RawHeader(MAuthRequest.MCC_AUTHENTICATION_HEADER_NAME, authHeaderV2),
+        RawHeader(MAuthRequest.X_MWS_TIME_HEADER_NAME, timeHeaderV2.toString),
+        RawHeader(MAuthRequest.X_MWS_AUTHENTICATION_HEADER_NAME, authHeader)
```
dgeyer-mdsol: Maybe there should be a separate test for  V2 headers and V1 headers

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
```diff
@@ -139,7 +178,205 @@ class MAuthDirectivesSpec extends WordSpec with Matchers with ScalatestRouteTest
...
+    "reject if request validation timeout passed" in {
+      (client.getPublicKey _).expects(*).never
+      //noinspection ConvertibleToMethodValue
+      (mockEpochTimeProvider.inSeconds _: () => Long).expects().returns(timeHeader)
+
+      Get().withHeaders(
+        RawHeader(MAuthRequest.MCC_TIME_HEADER_NAME, (timeHeader - requestValidationTimeout.toSeconds - 10).toString),
+        RawHeader(MAuthRequest.MCC_AUTHENTICATION_HEADER_NAME, authHeaderV2),
+        RawHeader(MAuthRequest.X_MWS_TIME_HEADER_NAME, (timeHeader - requestValidationTimeout.toSeconds - 10).toString),
+        RawHeader(MAuthRequest.X_MWS_AUTHENTICATION_HEADER_NAME, authHeader)
```
dgeyer-mdsol: same here

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
```diff
@@ -139,7 +178,205 @@ class MAuthDirectivesSpec extends WordSpec with Matchers with ScalatestRouteTest
...
+    "reject with a MalformedHeaderRejection if supplied with bad format" in {
+      Get().withHeaders(RawHeader(MAuthRequest.X_MWS_TIME_HEADER_NAME, "xyz")) ~> route ~> check {
+        inside(rejection) { case MalformedHeaderRejection("x-mws-time", "x-mws-time header supplied with bad format: [xyz]", None) ⇒ }
+      }
+    }
+
+    "reject with a MalformedHeaderRejection if supplied with bad format for V2" in {
+      Get().withHeaders(
+        RawHeader(MAuthRequest.MCC_TIME_HEADER_NAME, "xyz"),
+        RawHeader(MAuthRequest.X_MWS_TIME_HEADER_NAME, "1234567")
```
dgeyer-mdsol: why is this passed in this test?

modules/mauth-authenticator/src/main/scala/com/mdsol/mauth/scaladsl/RequestAuthenticator.scala
```diff
@@ -39,29 +48,73 @@ class RequestAuthenticator(publicKeyProvider: ClientPublicKeyProvider, epochTime
...
+    )
+
+    // Compare the decrypted signature and the recreated signature hashes.
+    try {
+      MAuthSignatureHelper.verifyRSA(unencryptedRequestString, mAuthRequest.getRequestSignature, clientPublicKey)
+    } catch {
+      case ex: Exception =>
+        val message = "MAuth request validation failed because of " + ex.getMessage
+        logger.error(message)
+        throw new MAuthValidationException(message)
```
dgeyer-mdsol: Instead of throwing an exception this method could return an Either, or just logging an returning false?

modules/mauth-authenticator/src/test/scala/com/mdsol/mauth/scaladsl/utils/RequestAuthenticatorSpec.scala
```diff
@@ -72,6 +72,47 @@ class RequestAuthenticatorSpec extends FlatSpec with RequestAuthenticatorBaseSpe
...
+      validationResult shouldBe true
+    }
+  }
+
+  it should "authenticate a valid request with V2 headers if disabled V1" in clientContext { (client) =>
+    //noinspection ConvertibleToMethodValue
+    (mockEpochTimeProvider.inSeconds _: () => Long).expects().returns(CLIENT_X_MWS_TIME_HEADER_VALUE.toLong + 3)
+    val authenticator = new RequestAuthenticator(client, mockEpochTimeProvider, true)
+
+    whenReady(authenticator.authenticate(getSimpleRequestV2)) { validationResult =>
```
dgeyer-mdsol: just for you to know, this could be:
`val result = authenticator.authenticate(getSimpleRequestV2).futureValue`
`result shouldBe true`

modules/mauth-authenticator/src/test/scala/com/mdsol/mauth/scaladsl/utils/RequestAuthenticatorSpec.scala
```diff
@@ -72,6 +72,47 @@ class RequestAuthenticatorSpec extends FlatSpec with RequestAuthenticatorBaseSpe
...
+      validationResult shouldBe true
+    }
+  }
+
+  it should "authenticate a valid request with all headers" in clientContext { (client) =>
+    //noinspection ConvertibleToMethodValue
```
dgeyer-mdsol: what is this for?

modules/mauth-authenticator/src/test/scala/com/mdsol/mauth/scaladsl/utils/RequestAuthenticatorSpec.scala
```diff
@@ -72,6 +72,47 @@ class RequestAuthenticatorSpec extends FlatSpec with RequestAuthenticatorBaseSpe
...
+    (mockEpochTimeProvider.inSeconds _: () => Long).expects().returns(CLIENT_X_MWS_TIME_HEADER_VALUE.toLong + 3)
+    val authenticator = new RequestAuthenticator(client, mockEpochTimeProvider)
+
+    whenReady(authenticator.authenticate(getRequestWithAllHeaders)) { validationResult =>
+      validationResult shouldBe true
+    }
+  }
+
+  it should "not authenticate a request V1 header only if disabled V1" in {
```
dgeyer-mdsol: maybe this description reads better as: should reject a request with V1 headers when V1 is disabled

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
bgong-mdsol: This test is to verify the client works well with the both headers of V1 and V2 provided

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
bgong-mdsol: The same as above

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
bgong-mdsol: The client always pick up V2 header MCC_TIME_HEADER_NAME if the V2 header is provided, the time header should be number, otherwise it throws MalformedHeaderRejection

modules/mauth-authenticator/src/test/scala/com/mdsol/mauth/scaladsl/utils/RequestAuthenticatorSpec.scala
bgong-mdsol: This is for verify with the both v1 and v2 headers provided. I have renamed as "authenticate a valid request with the both V1 and V2 headers provided"

modules/mauth-authenticator/src/test/scala/com/mdsol/mauth/scaladsl/utils/RequestAuthenticatorSpec.scala
bgong-mdsol: updated as suggested

modules/mauth-authenticator/src/test/scala/com/mdsol/mauth/scaladsl/utils/RequestAuthenticatorSpec.scala
dgeyer-mdsol: I meant the comment :)

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
dgeyer-mdsol: Oh, I see, it is because the signer returns the two versions of headers, do you think it worth to test all the possible valid cases,   both V1 and V2 passed, only V1 and only V2?

modules/mauth-authenticator/src/main/scala/com/mdsol/mauth/scaladsl/RequestAuthenticator.scala
bgong-mdsol: changes as suggested( just logging and returning false)

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
dgeyer-mdsol: I mean, this test will pass if X_MWS_TIME_HEADER_NAME is not passed, so, why passing it?

bgong-mdsol: We like to check the client will read and check the V2 header MCC_TIME_HEADER_NAME only, ignore the V1 header passed if MCC_TIME_HEADER_NAME is passed.

modules/mauth-authenticator/src/test/scala/com/mdsol/mauth/scaladsl/utils/RequestAuthenticatorSpec.scala
bgong-mdsol: updated a suggested

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
bgong-mdsol: added more  valid cases as suggested

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
```diff
@@ -269,71 +218,66 @@ trait MAuthDirectives extends StrictLogging {
...
-  def extractLatestAuthenticationHeaders(request: HttpRequest, disableV1: Boolean): Directive1[MauthHeaderValues] = {
-    extractAuthenticationHeaders(request, `MCC-Authentication`.name, `MCC-Time`.name, false).flatMap { mauthHeader: MauthHeaderValues =>
-      {
-        if (!mauthHeader.authenticator.isEmpty) {
-          provide(mauthHeader)
+  def extractLatestAuthenticationHeaders(disableV1: Boolean): Directive1[MauthHeaderValues] = {
+    extractRequest.flatMap { request: HttpRequest =>
+      // Try to extract and verify V2 headers
+      val authenticator = extractRequestHeader(request, `MCC-Authentication`.name)
+      if (!authenticator.isEmpty) {
```
jatcwang: * can use `nonEmpty`
* Should this name be something like `authenticationHeaderStr`?

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
```diff
@@ -269,71 +218,66 @@ trait MAuthDirectives extends StrictLogging {
...
+              logger.error(msg)
+              reject(MalformedHeaderRejection(headerName = `MCC-Time`.name, errorMsg = msg, None))
+            }
+          }
+        } else {
+          reject(MissingHeaderRejection(`MCC-Time`.name))
+        }
+      } else {
+        // If V2 headers not found, fallback to V1 headers if allowed
+        if (!disableV1) {
```
jatcwang: Let's avoid double negatives in our code. How about `enableV1`?

jatcwang:
Thanks this is much cleaner now since the extraction logic is all in one place.

I don't think we need classes like `MCC-Authentication` anymore (they should just be constants, really) - can they be removed?

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
```diff
@@ -88,16 +107,33 @@ trait MAuthDirectives extends StrictLogging {
...
+  val extractMCCAuthenticationDetail: Directive1[AuthHeaderDetail] =
+    headerValueByType[`MCC-Authentication`]((): Unit).flatMap { hdr =>
+      extractAuthHeaderDetail(hdr.value, MAuthVersion.MWSV2) match {
+        case Some(ahd: AuthHeaderDetail) => provide(ahd)
+        case None =>
+          val msg = s"${`MCC-Authentication`.name} header supplied with bad format: [${hdr.value}]"
+          logger.error(msg)
+          reject(MalformedHeaderRejection(headerName = `MCC-Authentication`.name, errorMsg = msg, None))
+      }
+    }
```
jatcwang: I don't think we need these directives. If we delete them we can delete the likes of `MCC-Authentication`. (old values like `extractMwsTimeHeader` will have to be kept for binary compatibility reasons)

jatcwang: we should deprecate directives like `extractMwsTimeHeader` too. User should be using `MAuthDirectives#authenticate` to avoid issues and get bug fixes automatically

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
```diff
@@ -88,16 +107,33 @@ trait MAuthDirectives extends StrictLogging {
...
   /////////////////////////////////////////////
   //  Utility functions
   /////////////////////////////////////////////
-
-  private def extractAuthHeaderDetail(str: String): Option[AuthHeaderDetail] = {
-    if (str.startsWith("MWS ")) {
-      str.replaceFirst("MWS ", "").split(":").toList match {
+  private def extractAuthHeaderDetail(str: String, mauthVersion: MAuthVersion): Option[AuthHeaderDetail] = {
+    val regexString = mauthVersion.getValue + " "
+    if (str.startsWith(regexString)) {
```
jatcwang: this is not a regex. It's a constant

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
```diff
@@ -112,6 +148,74 @@ trait MAuthDirectives extends StrictLogging {
...
     }
   }
+
+  private def getQueryString(req: HttpRequest): String = req.uri.rawQueryString.getOrElse("")
+
+  private def extractRequestHeader(request: HttpRequest, headerName: String): String =
+    request.headers
+      .find(_.name().compareToIgnoreCase(headerName) == 0)
+      .map(_.value())
+      .getOrElse("")
```
jatcwang: `request.getHeader(headerName).map(_.value()).orElse("")`

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
```diff
@@ -160,20 +188,36 @@ public Builder withQueryParameters(String queryParameters) {
...
-          timeHeaderValue = headers.get(MCC_TIME_HEADER_NAME);
+      if (mauthHeaders != null && !mauthHeaders.isEmpty()) {
+        if (mauthHeaders.get(MCC_AUTHENTICATION_HEADER_NAME) != null) {
+          authenticationHeaderValue = mauthHeaders.get(MCC_AUTHENTICATION_HEADER_NAME);
+          timeHeaderValue = mauthHeaders.get(MCC_TIME_HEADER_NAME);
         } else {
-          authenticationHeaderValue = headers.get(X_MWS_AUTHENTICATION_HEADER_NAME);
-          timeHeaderValue = headers.get(X_MWS_TIME_HEADER_NAME);
+          authenticationHeaderValue = mauthHeaders.get(X_MWS_AUTHENTICATION_HEADER_NAME);
+          timeHeaderValue = mauthHeaders.get(X_MWS_TIME_HEADER_NAME);
```
jatcwang: Header extraction logic needs to be case insensitive.

modules/mauth-common/src/main/java/com/mdsol/mauth/util/MAuthSignatureHelper.java
```diff
@@ -35,18 +34,18 @@
...
-   * @param requestBody: request body string
+   * @param requestBody: request body in byte[]
    * @param epochTime: current seconds since Epoch
    * @param mauthVersion: Mauth version (MSW or MSWV2)
    * @return String
    *
    * @throws MAuthSigningException
    */
   public static String generateStringToSign(UUID appUUID, String httpMethod, String resourceUrl, String queryParameters,
-      String requestBody, String epochTime, MAuthVersion mauthVersion) throws MAuthSigningException {
```
jatcwang: This breaks binary compatibility. Should leave this method alone (with deprecation)

modules/mauth-common/src/main/java/com/mdsol/mauth/util/MAuthSignatureHelper.java
```diff
@@ -164,10 +163,15 @@ public static String encryptSignature(PrivateKey privateKey, String unencryptedS
...
+
+  public static String getHexEncodedDigestedString(byte[] unencryptedData) {
     try {
       // Get digest
       MessageDigest md = MessageDigest.getInstance("SHA-512");
-      byte[] digestedString = md.digest(unencryptedString.getBytes(StandardCharsets.UTF_8));
+      byte[] digestedString = md.digest(unencryptedData);
       // Convert to hex
       return Hex.encodeHexString(digestedString);
     } catch (NoSuchAlgorithmException ex) {
```
jatcwang: (line below)

alghoritm -> algorithm

modules/mauth-proxy/src/main/resources/reference.conf
```diff
@@ -8,4 +8,8 @@ proxy {
 app {
   uuid: ${?APP_MAUTH_UUID}
   private_key: ${?APP_MAUTH_PRIVATE_KEY}
+}
+
+mauth {
+  disable_v1: false
```
jatcwang: should this be `enable_v1 : true`?

modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/MAuthRequestSigner.scala
```diff
@@ -7,35 +7,42 @@ import java.util.UUID
...
 
 case class CryptoError(msg: String, cause: Option[Throwable] = None)
 
 trait RequestSigner {
   def signRequest(request: UnsignedRequest): Either[Throwable, SignedRequest]
 }
 
-class MAuthRequestSigner(appUUID: UUID, privateKey: PrivateKey, epochTimeProvider: EpochTimeProvider)
-    extends DefaultSigner(appUUID, privateKey, epochTimeProvider)
+class MAuthRequestSigner(appUUID: UUID, privateKey: PrivateKey, epochTimeProvider: EpochTimeProvider, disableV1: Boolean)
```
jatcwang: enableV1?

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
```diff
@@ -152,7 +152,7 @@ private boolean validateSignatureV2 (MAuthRequest mAuthRequest, PublicKey client
         MAuthSignatureHelper.generateStringToSign(
             mAuthRequest.getAppUUID(), mAuthRequest.getHttpMethod(), mAuthRequest.getResourcePath(),
             mAuthRequest.getQueryParameters(),
-            new String(mAuthRequest.getMessagePayload(), StandardCharsets.UTF_8),
+            mAuthRequest.getMessagePayload(),
             String.valueOf(mAuthRequest.getRequestTime()),
```
jatcwang: If this is always a `long` the function should take a `long` instead of `String` and move the long -> String conversion into `generateStringToSign`. The general principle is to make it hard for users of your API to screw things up

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
bgong-mdsol: updated as suggested

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
bgong-mdsol: updated as suggested

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
bgong-mdsol: renamed

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
bgong-mdsol: updated as suggested

modules/mauth-common/src/main/java/com/mdsol/mauth/util/MAuthSignatureHelper.java
bgong-mdsol: fixed

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
bgong-mdsol: removed directives  of MCC-Authentication, MCC-Time

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
bgong-mdsol: mauthHeaders is Map<String, String> to store mauth headers provided by consumer and have mentioned the supported mauth headers in readme for Java user. I can change it to new TreeMap<String, String>(String.CASE_INSENSITIVE_ORDER)

modules/mauth-common/src/main/java/com/mdsol/mauth/util/MAuthSignatureHelper.java
bgong-mdsol: **generateStringToSign** is a new method to support the both V1 and V2.  **generateUnencryptedSignature** is the original one for V1 and @Deprecated already

bgong-mdsol: Regarding requestBody is  an array of bytes,  the implementation 
 of **generateUnencryptedSignature** doesn't support non-UTF-8 string and I just leave as it is. But we do support binary body for V2 in **generateStringToSignV2**

modules/mauth-proxy/src/main/resources/reference.conf
bgong-mdsol: Currently jvm version does the same way as .net version. will re-think this. Thanks

modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/MAuthRequestSigner.scala
bgong-mdsol: same as above

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
bgong-mdsol: It is just follow the original interface of **generateUnencryptedSignature**, easy for the current user do upgrade?

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
jatcwang: yeap sounds good. Make `withMauthHeaders` take a Map<String, String> and insert it into `this.mauthHeaders` which is a TreeMap

modules/mauth-common/src/main/java/com/mdsol/mauth/util/MAuthSignatureHelper.java
jatcwang: The public signature of `generateStringToSign` is still broken though..I can pass a non UTF-8 byte array with MWS V1 and it'll blow up. Looking at the usages, all call sites already know what the MWS version is, so do we need this helper function at all?

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
jatcwang: sure I guess this interface is also easy when you want to validate an incoming request (since all headers are strings)

modules/mauth-common/src/main/java/com/mdsol/mauth/util/MAuthSignatureHelper.java
bgong-mdsol: Yeah, if it is V1, generateStringToSign calls generateUnencryptedSignature that is existing method for V1 and doesn't  support non UTF-8 byte array.
It seems some Java modules are using this function.

jatcwang:
There are still some `disable_v1` references (unused variables etc). Can we remove them?

modules/mauth-common/src/main/java/com/mdsol/mauth/util/MAuthSignatureHelper.java
jatcwang: Yes that's my point - we're introducing a potential issue here with the API as it is now.

Let's look at all usages of `generateStringToSign`

* com.mdsol.mauth.RequestAuthenticator#validateSignatureV2: hardcoded to V2
* com.mdsol.mauth.scaladsl.RequestAuthenticator#validateSignatureV2: hardcoded to V2
* com.mdsol.mauth.DefaultSigner#generateRequestHeadersV2: hardcoded to V2

So there's no need for generateStringToSign at all. Any of those use sites could've call `generateStringToSignV2` directly.

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
jatcwang: Hmm after annotating the type it works for me: `request.getHeader(headerName).map[String](_.value()).orElse("")`

bgong-mdsol: Yeah,  this works 👍

bgong-mdsol:
> There are still some `disable_v1` references (unused variables etc). Can we remove them?

Removed or reworded "disable_v1" https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/dc4c71dcb26529cf7715860ff77791c74c51713c

modules/mauth-common/src/main/java/com/mdsol/mauth/util/MAuthSignatureHelper.java
bgong-mdsol: removed generateStringToSign  as suggested. See https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/1105451ea7c6fdbab0893eae57da45edac6d2021

modules/mauth-authenticator-akka-http/src/main/scala/com/mdsol/mauth/akka/http/MAuthDirectives.scala
```diff
@@ -91,13 +110,13 @@ trait MAuthDirectives extends StrictLogging {
   /////////////////////////////////////////////
   //  Utility functions
   /////////////////////////////////////////////
-
-  private def extractAuthHeaderDetail(str: String): Option[AuthHeaderDetail] = {
-    if (str.startsWith("MWS ")) {
-      str.replaceFirst("MWS ", "").split(":").toList match {
+  private def extractAuthHeaderDetail(str: String, mauthVersion: MAuthVersion): Option[AuthHeaderDetail] = {
+    val mauthVersionStr = mauthVersion.getValue + " "
+    if (str.startsWith(mauthVersionStr)) {
```
jatcwang: We can revert change for this function because it is only used for v1. 
(The extraction logic is buggy too, because a v2 header value starts with "MWS" too)

You can rename or inline this method - it is private so there's no binary compatibility concern

bgong-mdsol: Acturaly, it is checking the value starts with mauthVersion.getValue + " " ("MWS ").
Anyway, I reverted the changes of AuthHeaderDetail, extractMAuthHeader and extractAuthHeaderDetail since V2 doesn't need these. See https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/41ff1585613b4d6b13840f5e0a6624440325b903

modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/MAuthRequestSigner.scala
```diff
@@ -7,35 +7,42 @@ import java.util.UUID
...
+  * @note it includes V2 headers only if V2 only is enabled, otherwise it includes the both V1 and V2 headers
   *
-  * @param req        The original request that was used to create this object
-  * @param authHeader The Auth header information
-  * @param timeHeader The Time header information
+  * @param req          The original request that was used to create this object
+  * @param authHeaders  The Auth header information
   */
-case class SignedRequest(req: UnsignedRequest, authHeader: String, timeHeader: String)
+case class SignedRequest(req: UnsignedRequest, authHeaders: Map[String, String])
```
jatcwang: This breaks binary compatibility

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
```diff
@@ -35,11 +35,22 @@ class MAuthDirectivesSpec extends WordSpec with Matchers with ScalatestRouteTest
...
+       |T3apKHcV/6cJ9asm+jDjzB8OuCVWVsLZQKQbtiydUYNisYerKVxWPLs9SHNZ6GmAqq4ZCCpyEQZuMNF6cMmXgQ0Pxe9
+       |X/yNA1Xc3Fakuga47lUQ6Bn7xvhkH6P+ZP0k4U7kidziXpxpkDts8fEXTpkvFX0PR7vaxjbMZzWsU413jyNsw==;""".stripMargin.replaceAll("\n", "")
+  private val authHeaderV2: String = s"$authPrefixV2 $appUuid:$signatureV2"
+
   private implicit val timeout: FiniteDuration = 30 seconds
   private implicit val requestValidationTimeout: Duration = 10 seconds
   private val client = mock[ClientPublicKeyProvider]
   private val mockEpochTimeProvider: EpochTimeProvider = mock[EpochTimeProvider]
   private implicit val authenticator: RequestAuthenticator = new RequestAuthenticator(client, mockEpochTimeProvider)
+  private implicit val authenticatorV2: RequestAuthenticator = new RequestAuthenticator(client, mockEpochTimeProvider, true)
```
jatcwang: Use named parameters please :)

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
```diff
@@ -184,4 +196,229 @@ class MAuthDirectivesSpec extends WordSpec with Matchers with ScalatestRouteTest
...
+      //noinspection ConvertibleToMethodValue
+      (mockEpochTimeProvider.inSeconds _: () => Long).expects().returns(timeHeader)
+
+      Get().withHeaders(
+        RawHeader(MAuthRequest.MCC_TIME_HEADER_NAME, (timeHeader - requestValidationTimeout.toSeconds - 10).toString),
+        RawHeader(MAuthRequest.MCC_AUTHENTICATION_HEADER_NAME, authHeaderV2),
+        RawHeader(MAuthRequest.X_MWS_TIME_HEADER_NAME, (timeHeader - requestValidationTimeout.toSeconds - 10).toString),
+        RawHeader(MAuthRequest.X_MWS_AUTHENTICATION_HEADER_NAME, authHeader)
+      ) ~> route ~> check {
+        inside(rejection) { case MdsolAuthFailedRejection ⇒ }
```
jatcwang: Can you update these unicode character to be `=>`?

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
```diff
@@ -184,4 +196,229 @@ class MAuthDirectivesSpec extends WordSpec with Matchers with ScalatestRouteTest
       }
     }
   }
+
+  "authenticate for V2" should {
+    lazy val route: Route = authenticate(executor, authenticatorV2, timeout, requestValidationTimeout).apply(complete(HttpResponse()))
+    val publicKey = MAuthKeysHelper.getPublicKeyFromString(FixturesLoader.getPublicKey)
+
+    "pass successfully authenticated request with parameters" in {
```
jatcwang: "...with both V1 and V2 headers ,with V2 headers taking precedence"

To test precedence works, please use invalid values for v1 headers.

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
```diff
@@ -184,4 +196,229 @@ class MAuthDirectivesSpec extends WordSpec with Matchers with ScalatestRouteTest
       }
     }
   }
+
+  "authenticate for V2" should {
```
jatcwang: Should we add test cases for the following?

* header name detection should be case insensitive
* The default RequestAuthenticator (where `v2OnlyAuthenticate = false`) will extract/handle both v1 and v2 headers, with v2 headers taking precedence. This should be under the `"authenticate"` test section

jatcwang: Rename this to be "authenticate when v2OnlyAuthenticate = true" for clarity

modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/http/MAuthHeaders.scala
```diff
@@ -36,3 +36,35 @@ object `X-MWS-Time` extends ModeledCustomHeaderCompanion[`X-MWS-Time`] {
 
   override def parse(value: String): Try[`X-MWS-Time`] = Try(new `X-MWS-Time`(value))
 }
+
+final class `MCC-Authentication`(token: String) extends ModeledCustomHeader[`MCC-Authentication`] {
```
jatcwang: No longer needed. Same for `MCC-Time`

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
bgong-mdsol: Updated and added a couple specs to verify https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/735bd0daeb6afd8ec05d1562413a7721f033797f

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
bgong-mdsol: updated as suggested https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/735bd0daeb6afd8ec05d1562413a7721f033797f

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
bgong-mdsol: fixed https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/735bd0daeb6afd8ec05d1562413a7721f033797f

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
bgong-mdsol: updated as suggested https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/735bd0daeb6afd8ec05d1562413a7721f033797f

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
bgong-mdsol: added and updated as suggested https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/735bd0daeb6afd8ec05d1562413a7721f033797f

bgong-mdsol: updated ad suggested https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/735bd0daeb6afd8ec05d1562413a7721f033797f

modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/http/MAuthHeaders.scala
bgong-mdsol: removed the both MCC-Authentication and MCC-Time

modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/MAuthRequestSigner.scala
bgong-mdsol: Updated SignedRequest https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/f09b5f748be5b11e107ec3296906f97e2f3eadb5

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
```diff
@@ -50,7 +50,8 @@ class MAuthDirectivesSpec extends WordSpec with Matchers with ScalatestRouteTest
   private val client = mock[ClientPublicKeyProvider]
   private val mockEpochTimeProvider: EpochTimeProvider = mock[EpochTimeProvider]
   private implicit val authenticator: RequestAuthenticator = new RequestAuthenticator(client, mockEpochTimeProvider)
-  private implicit val authenticatorV2: RequestAuthenticator = new RequestAuthenticator(client, mockEpochTimeProvider, true)
+  private val v2OnlyAuthenticate = true
+  private implicit val authenticatorV2: RequestAuthenticator = new RequestAuthenticator(client, mockEpochTimeProvider, v2OnlyAuthenticate)
```
jatcwang: In principle this doesn't help you make sure that you're passing in the right parameters.

```
def abc(a: Boolean, b: Boolean) = ...

val a = true
val b = false
abc(b, a) // bug!
```

In Scala you have named parameters.

```
abc(a = true, b = false)
```

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
```diff
@@ -191,31 +206,16 @@ class MAuthDirectivesSpec extends WordSpec with Matchers with ScalatestRouteTest
...
-      Get("/").withHeaders(
-        RawHeader(MAuthRequest.MCC_TIME_HEADER_NAME, timeHeader.toString),
-        RawHeader(MAuthRequest.MCC_AUTHENTICATION_HEADER_NAME, authHeaderV2),
-        RawHeader(MAuthRequest.X_MWS_TIME_HEADER_NAME, timeHeaderV2.toString),
-        RawHeader(MAuthRequest.X_MWS_AUTHENTICATION_HEADER_NAME, authHeader)
-      ) ~> route ~> check {
-        status shouldEqual StatusCodes.OK
-      }
-    }
-
```
jatcwang: Why was this test removed? This test should still be successful no?

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
jatcwang: Nicely done

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
jatcwang: btw this doesn't need to be an implicit (same for `authenticator`)

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
```diff
@@ -179,9 +206,233 @@ class MAuthDirectivesSpec extends WordSpec with Matchers with ScalatestRouteTest
...
+      //noinspection ConvertibleToMethodValue
+      (mockEpochTimeProvider.inSeconds _: () => Long).expects().returns(timeHeader)
+
+      Get("/").withHeaders(
+        RawHeader(MAuthRequest.MCC_TIME_HEADER_NAME, timeHeader.toString),
+        RawHeader(MAuthRequest.MCC_AUTHENTICATION_HEADER_NAME, authHeaderV2)
+      ) ~> route ~> check {
+        status shouldEqual StatusCodes.OK
+      }
+    }
```
jatcwang: I think we're still missing case-insensitive header extraction test.

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
bgong-mdsol: updated as suggested https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/07fa8b14f1df3b53cbccf169c0db9217561879bf

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
bgong-mdsol: I moved the test into "authenticate" and added it back for "authenticate when v2OnlyAuthenticate = true". the both should be be successful.  https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/07fa8b14f1df3b53cbccf169c0db9217561879bf

modules/mauth-authenticator-akka-http/src/test/scala/com/mdsol/mauth/akka/http/MAuthDirectivesSpec.scala
bgong-mdsol: added a test with the headers in uppercase https://github.com/mdsol/mauth-jvm-clients/pull/92/commits/07fa8b14f1df3b53cbccf169c0db9217561879bf

modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/MAuthRequestSigner.scala
jatcwang: It still isn't binary compatible unfortunately. Let's address this in another PR.

jfeltesse-mdsol:
Thanks @jatcwang for all the input! Highly appreciated.

Issue #65, opened 2018-07-24
Need to update mauth-client to include query parameters in mAuth signature


raroskar-mdsol:
Background
Medicines and Healthcare products Regulatory Agency (MHRA)'s review of Medidata Products, they observed, **
The protocol used to authenticate remote applications, mAuth, was used inconsistently in that important elements of the message were excluded from the signature (specifically the use of function verbs in URL queries), thus functions were not authenticated, which could introduce the risk that functions could be substituted (e.g. overwrite data instead of append to a record). **

What is expected?

- mauth-client has new mauth request signing algorithm (MWSV2) , which will have authentication header looking similar to original 'x-mws-authentication' header and in addition it will append the query params k-v pairs, sorted ascending by key.
- mauth-client should maintain backwards compatibility i.e. original MWS (current signature algorithm) should still be available and usable.
- mauth-client must be able to sign with both MWS and MWSV2 for requests.
- mauth-client should have sufficient logging such that it will be easy to figure out how many (and who) requests are still using MWS vs MWSV2 in any given time period

jatcwang:
Most of these are addressed in https://github.com/mdsol/mauth-jvm-clients/pull/94

@bgong-mdsol After the current Pr is merged should we add logging when a mauth v1 only request is received? (Addressing the last point)

bgong-mdsol:
> Most of these are addressed in #94
> 
> @bgong-mdsol After the current Pr is merged should we add logging when a mauth v1 only request is received? (Addressing the last point)

@jatcwang we add logging for the both  in https://github.com/mdsol/mauth-jvm-clients/blob/72583e5ccd2d10fefc747111930732c5adc49aad/modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java#L93

jatcwang:
What's the log message if both v1 and v2 headers are provided?

bgong-mdsol:
> What's the log message if both v1 and v2 headers are provided?
> 
> @jatcwang Yes, add the mauth version (MWS or MWSV2) in the logging https://github.com/mdsol/mauth-jvm-clients/blob/72583e5ccd2d10fefc747111930732c5adc49aad/modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java#L93~#L95

If both v1 and v2 headers are provided, Authenticator does authenticate for V2 (higher version as design), only V2 related information will be wrote into log.

[...]

PR #96, opened 2020-01-03
[MCC-569318]Updated authenticator for v1 to support binary payload

bgong-mdsol:
**Background**
The existing Authenticator of mauth java client doesn't support binary payload for Mauth V1. Some applications (e.g. RCM-API) need to authenticate the request with binary payload. Because of this, RCM has its own mauth implementation

**Changes Summary**
- add a method generateUnencryptedSignature to export a byte array in mauth-common;
- enhance validateSignatureV1 in mauth-authenticator to support binary payload;
- extend test fixtures to test binary payload (use the same as Ruby and Python);
- add the test case for the request with binary payload;


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "2.0.0-SNAPSHOT"
```
jatcwang: 2.0.0 hasn't been released yet I think? (We still need to address https://github.com/mdsol/mauth-jvm-clients/issues/65#issuecomment-566651734)

modules/mauth-common/src/main/java/com/mdsol/mauth/util/MAuthSignatureHelper.java
```diff
@@ -51,6 +49,35 @@ public static String generateUnencryptedSignature(UUID appUUID, String httpMetho
...
+   *   httpMethod + "\n" + resourceUrl + "\n" + requestBody + "\n" + app_uuid + "\n" + epochTime
+   *
+   */
+  @Deprecated
+  public static byte[] generateUnencryptedSignature(UUID appUUID, String httpMethod, String resourceUrl, byte[] requestBody, String epochTime) throws IOException{
+    logger.debug("Generating byte[] to sign for V1");
+    ByteArrayOutputStream baos = new ByteArrayOutputStream();
+    String part1 = httpMethod + "\n" + resourceUrl + "\n";
+    String part2 =  "\n" + appUUID.toString() + "\n" + epochTime;
+    baos.write(part1.getBytes());
```
jatcwang: When converting String to byte[], you should always specify the encoding

bgong-mdsol: fixed, use UTF_8

modules/mauth-test-utils/src/main/resources/keys/fake_publickey2.pem
```diff
@@ -0,0 +1,9 @@
+-----BEGIN PUBLIC KEY-----
+MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAs3zgudWufUW/UFoQtxfB
+ePRDqc63RdpvuBILKHNFAYpxxuSRMzQUKSrsvXpWf983/kTtvXBrj4Hm6qWC9fqG
+Kh/qDTnjamXUzxg6utFbcrOrc4hsfvfRjEaymclTPLXRa9+iD3vTzDBqS85eDFDm
+5hTpPwc3C/7+iw0cfWqwP7jJsn5Ovul50UOuR+czeBWuQXZPh21nUpEI7ZW8U3W2
+lmO2DrSez2FVtW3Oiek/AQz+uOtlNSPoJcH06O2dlh7AAF8c3eFTaYH4gBTFzcBU
+p3BneZ2cr0e5m5eVRcCOFqDGQF3YJQT2I1EpMecRWo7UW48mF0wzVj/mK1rQUAeb
+gQIDAQAB
+-----END PUBLIC KEY-----
```
jatcwang: Any reason why we need another public key?

version.sbt
bgong-mdsol: Rollback to 2.0.0-SNAPSHOT.

modules/mauth-authenticator/src/test/scala/com/mdsol/mauth/RequestAuthenticatorSpec.scala
```diff
@@ -87,4 +89,13 @@ class RequestAuthenticatorSpec extends AnyFlatSpec with RequestAuthenticatorBase
...
 
+  it should "validate a valid request with binary body for V1" in {
+    //noinspection ConvertibleToMethodValue
+    (mockEpochTimeProvider.inSeconds _: () => Long).expects().returns(CLIENT_X_MWS_TIME_HEADER_BINARY_VALUE.toLong + 3)
+    (mockClientPublicKeyProvider.getPublicKey _)
+      .expects(UUID.fromString(CLIENT_REQUEST_BINARY_APP_UUID))
+      .returns(MAuthKeysHelper.getPublicKeyFromString(PUBLIC_KEY2))
+    authenticator.authenticate(getRequestWithBinaryBodyV1) shouldBe true
+  }
+
```
jatcwang: Probably a good idea to check that the string-body API and byte-array API will both succeed (they should be equivalent if their string body is the same)

modules/mauth-test-utils/src/main/resources/keys/fake_publickey2.pem
bgong-mdsol: We try to sync up the test fixture with other platform (Ruby version is standard).  I use the same test data as Ruby and Python in https://github.com/mdsol/mauth-jvm-clients/blob/384557bc88e9958b3295049bff5b310494ccd739/modules/mauth-authenticator/src/test/scala/com/mdsol/mauth/RequestAuthenticatorBaseSpec.scala#L140-L151
The new public key (fake_publickey2.pem) is for the private key which is used to generate the signature  CLIENT_REQUEST_SIGNATURE_BINARY_V1 https://github.com/mdsol/mauth-jvm-clients/blob/384557bc88e9958b3295049bff5b310494ccd739/modules/mauth-authenticator/src/test/scala/com/mdsol/mauth/RequestAuthenticatorBaseSpec.scala#L144-L150

jatcwang: Fair enough

modules/mauth-authenticator/src/test/scala/com/mdsol/mauth/RequestAuthenticatorSpec.scala
bgong-mdsol: This is a test for the request with string payload
https://github.com/mdsol/mauth-jvm-clients/blob/384557bc88e9958b3295049bff5b310494ccd739/modules/mauth-authenticator/src/test/scala/com/mdsol/mauth/RequestAuthenticatorSpec.scala#L22

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
```diff
@@ -128,18 +128,23 @@ private boolean validateSignatureV1 (MAuthRequest mAuthRequest, PublicKey client
...
+          String.valueOf(mAuthRequest.getRequestTime())
+      );
+      messageDigest_bytes = MAuthSignatureHelper.getHexEncodedDigestedString(messageDigest_bytes).getBytes(StandardCharsets.UTF_8);
+
+      // Compare the decrypted signature and the recreated signature hashes.
+      // If both match, the request was signed by the requesting application and is valid.
+      return Arrays.equals(messageDigest_bytes, decryptedSignature);
+    } catch (Exception ex) {
+      final String message = "MAuth request validation failed because of " + ex.getMessage();
+      logger.error(message);
```
austek: why not pass throwable to logger?

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
```diff
@@ -128,18 +128,23 @@ private boolean validateSignatureV1 (MAuthRequest mAuthRequest, PublicKey client
...
+      throw new MAuthValidationException(message);
```
austek: probably adding `MAuthValidationException(String message, Throwable cause)` constructor and use that instead of passing just the message of the trowable would be better

bgong-mdsol: Updated in https://github.com/mdsol/mauth-jvm-clients/pull/96/commits/7de735546c1694e3da46b41c7a6ff4d0268c7005

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
bgong-mdsol: Updated in https://github.com/mdsol/mauth-jvm-clients/pull/96/commits/7de735546c1694e3da46b41c7a6ff4d0268c7005

modules/mauth-authenticator/src/test/scala/com/mdsol/mauth/RequestAuthenticatorSpec.scala
jatcwang: Ah sorry I see you changed the implementation of validateSignatureV1 to always use bytes[] body

jatcwang:
I think we need to change the implementation of `DefaultSigner#generateRequestHeaders`.

In `generateRequestHeaders`, it is given a byte[], but tries to create a String from it so it can call its own `DefaultSigner#generateRequestHeadersV1` which currently requires a string. This will therefore fail if `generateRequestHeaders` is provided with any non-UTF8 byte[]. 

We should change `DefaultSigner#generateRequestHeadersV1` to take a byte[] instead. Since `DefaultSigner#generateRequestHeadersV1` is `private`, you can modify its signature directly without breaking binary compatibility :)

bgong-mdsol:
> I think we need to change the implementation of `DefaultSigner#generateRequestHeaders`.
> 
> In `generateRequestHeaders`, it is given a byte[], but tries to create a String from it so it can call its own `DefaultSigner#generateRequestHeadersV1` which currently requires a string. This will therefore fail if `generateRequestHeaders` is provided with any non-UTF8 byte[].
> 
> We should change `DefaultSigner#generateRequestHeadersV1` to take a byte[] instead. Since `DefaultSigner#generateRequestHeadersV1` is `private`, you can modify its signature directly without breaking binary compatibility :)

Refactor DefaultSigner#generateRequestHeadersV1 and added specs for binary payload, see  https://github.com/mdsol/mauth-jvm-clients/pull/96/commits/69c9063468570b8a662c4324c7394b0e3f243267

jatcwang:
MAuth V1 spec does not specify how a request with binary body should be signed. Talked to @bgong-mdsol and she confirmed that the behaviour implemented in this PR (base64 the body) is what Ruby/Python implementation do. Thanks

PR #101, opened 2020-01-15
[MCC-530719] Fix to customize content-type in mAuthV2

mcallonalvarez:
#### What is this PR for?

Fix - Add support to customize content-type (apply "text/plain(UTF-8)" by default) in mAuthV2

#### What are the relevant JIRA tickets?
[MCC-530719](https://jira.mdsol.com/browse/MCC-530719)

This is also related to this previous bugfix https://github.com/mdsol/mauth-jvm-clients/pull/71

### DIFF
--- a/modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/http/Implicits.scala
+++ b/modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/http/Implicits.scala
@@ -4,8 +4,9 @@ import akka.http.scaladsl.model.headers.RawHeader
 import akka.http.scaladsl.model.{HttpEntity, _}
 import com.mdsol.mauth.SignedRequest
 import com.mdsol.mauth.http.HttpVerbOps._
-import com.mdsol.mauth.models.{UnsignedRequest => NewUnsignedRequest, SignedRequest => NewSignedRequest}
+import com.mdsol.mauth.models.{SignedRequest => NewSignedRequest}
 
+import scala.collection.immutable
 import scala.language.implicitConversions
 
 object Implicits {
@@ -32,7 +33,10 @@ object Implicits {
       * Create an akka-http request from a [[com.mdsol.mauth.models.SignedRequest]]
       */
     def toAkkaHttpRequest: HttpRequest = {
-      val allHeaders = (signedRequest.req.headers ++ signedRequest.mauthHeaders).toList
+      val contentType: Option[String] = extractContentTypeFromHeaders(signedRequest.req.headers)
+      val headersWithoutContentType: Map[String, String] = removeContentTypeFromHeaders(signedRequest.req.headers)
+
+      val allHeaders: immutable.Seq[RawHeader] = (headersWithoutContentType ++ signedRequest.mauthHeaders).toList
         .map {
           case (name, value) =>
             RawHeader(name, value)
@@ -41,7 +45,7 @@ object Implicits {
       HttpRequest(
         method = signedRequest.req.httpMethod,
         uri = Uri(signedRequest.req.uri.toString),
-        entity = HttpEntity(signedRequest.req.body),
+        entity = getHttpEntity(contentType, signedRequest.req.body),
         headers = allHeaders
       )
     }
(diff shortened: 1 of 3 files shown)


modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/http/Implicits.scala
```diff
@@ -59,6 +63,17 @@ object Implicits {
...
+  private def getHttpEntity(contentTypeOptional: Option[String], entityBody: Array[Byte]) = {
+    contentTypeOptional match {
+      case Some(contentType) =>
+        ContentType.parse(contentType) match {
+          case Right(parsedContentType) => HttpEntity(parsedContentType, entityBody)
+          case _ => HttpEntity(DEFAULT_CONTENT_TYPE, entityBody)
+        }
+      case None => HttpEntity(DEFAULT_CONTENT_TYPE, entityBody)
+    }
+  }
```
jatcwang: This seems overly complicated. Can we simply use the original headers? I'm playing around with this - will report back

cabbott:
@mdsol/architecture-enablement please 👀

modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/http/Implicits.scala
mcallonalvarez: We need to remove from the original headers the content-type and add it to the HttpEntity

modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/http/Implicits.scala
```diff
@@ -32,7 +33,10 @@ object Implicits {
       * Create an akka-http request from a [[com.mdsol.mauth.models.SignedRequest]]
       */
     def toAkkaHttpRequest: HttpRequest = {
-      val allHeaders = (signedRequest.req.headers ++ signedRequest.mauthHeaders).toList
+      val contentType: Option[String] = extractContentTypeFromHeaders(signedRequest.req.headers)
```
red-benabas: @mcallonalvarez good spot!

modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/http/Implicits.scala
```diff
@@ -32,7 +33,10 @@ object Implicits {
...
+      val headersWithoutContentType: Map[String, String] = removeContentTypeFromHeaders(signedRequest.req.headers)
```
jatcwang: Don't need this as akka-http will remove it anyway. (Will be good to add a comment though)

modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/http/Implicits.scala
jatcwang: You're right...Akka-http does not allow overwriting the headers 🙄

modules/mauth-signer-akka-http/src/main/scala/com/mdsol/mauth/http/Implicits.scala
jatcwang: Nevermind, probably less confusing to remove it outright

modules/mauth-signer-akka-http/src/test/scala/com/mdsol/mauth/http/ImplicitsTest.scala
```diff
@@ -135,20 +135,14 @@ class ImplicitsTest extends AnyWordSpec with Matchers {
           RawHeader(MAuthRequest.MCC_TIME_HEADER_NAME, "mcc-time-value"),
           RawHeader(MAuthRequest.MCC_AUTHENTICATION_HEADER_NAME, "mcc-authentication-value"),
           RawHeader(MAuthRequest.X_MWS_TIME_HEADER_NAME, "x-mws-time-value"),
-          RawHeader(MAuthRequest.X_MWS_AUTHENTICATION_HEADER_NAME, "x-mws-authentication-value"),
-          RawHeader("Content-Type", "application/json")
+          RawHeader(MAuthRequest.X_MWS_AUTHENTICATION_HEADER_NAME, "x-mws-authentication-value")
         ).toString()
```
jatcwang: Can we also add a wiremock test to ensure that the request has the right content-type header?

austek:
These should address #68 as well, no?

jatcwang:
Yes @austek. MAuth V2 no longer assumes that the body is a valid UTF-8 string.

modules/mauth-signer-akka-http/src/test/scala/com/mdsol/mauth/http/ImplicitsTest.scala
mcallonalvarez: Added wiremock tests to check that the default content type is text/plain UTF-8 and also, that the httpclient is sending the customized content-type FYI @austek

PR #105, opened 2020-01-20
[MCC-575954]fall back to V1 when V2 authentication fails

bgong-mdsol:
As discussed in this https://docs.google.com/document/d/1UsKJfZeDJb3cbvi0B-z-KMXHrNEGo-GKhXi2k4OnMxU/edit#heading=h.k5ec2aaqcyc0
and same as in ruby https://github.com/mdsol/mauth-client-ruby/pull/38, this PR updates the logic to fall back to V1 authentication when V2 authentication fails.

Please review:
@jatcwang @austek @mcallonalvarez @mdsol/architecture-enablement

### DIFF
--- a/modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
+++ b/modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
@@ -106,15 +106,21 @@ public class RequestAuthenticator implements Authenticator {
 
     PublicKey clientPublicKey = clientPublicKeyProvider.getPublicKey(mAuthRequest.getAppUUID());
     if (mAuthRequest.getMauthVersion().equals(MAuthVersion.MWSV2)) {
-      boolean isValidated = validateSignatureV2(mAuthRequest, clientPublicKey);
-      if (!isValidated && !v2OnlyAuthenticate) {
-        isValidated = fallbackValidateSignatureV1(mAuthRequest, clientPublicKey);
+      boolean v2IsValidated = validateSignatureV2(mAuthRequest, clientPublicKey);
+      if (v2OnlyAuthenticate) {
+        return v2IsValidated;
       }
-      return isValidated;
+      else if (v2IsValidated) {
+        return v2IsValidated;
+      }
+      else {
+        return fallbackValidateSignatureV1(mAuthRequest, clientPublicKey);
+      }
     }
     else {
       return validateSignatureV1(mAuthRequest, clientPublicKey);
     }
+
   }
 
   // Check epoch time is not older than specified interval.
--- a/modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
+++ b/modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
@@ -91,8 +91,6 @@ public class RequestAuthenticator implements Authenticator {
 
   @Override
   public boolean authenticate(MAuthRequest mAuthRequest) {
-    final String msgFormat = "Mauth-client attempting to authenticate request from app with mauth app uuid %s using version %s.";
-    logger.info(String.format(msgFormat, mAuthRequest.getAppUUID(), mAuthRequest.getMauthVersion().getValue()));
 
     if (!(validateTime(mAuthRequest.getRequestTime()))) {
       final String message = "MAuth request validation failed because of timeout " + requestValidationTimeoutSeconds + "s";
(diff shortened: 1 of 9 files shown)


modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
```diff
@@ -108,7 +106,11 @@ public boolean authenticate(MAuthRequest mAuthRequest) {
 
     PublicKey clientPublicKey = clientPublicKeyProvider.getPublicKey(mAuthRequest.getAppUUID());
     if (mAuthRequest.getMauthVersion().equals(MAuthVersion.MWSV2)) {
-      return validateSignatureV2(mAuthRequest, clientPublicKey);
+      boolean isValidated = validateSignatureV2(mAuthRequest, clientPublicKey);
+      if (!isValidated && !v2OnlyAuthenticate) {
+        isValidated = fallbackValidateSignatureV1(mAuthRequest, clientPublicKey);
+      }
+      return isValidated;
```
jatcwang: Please avoid boolean inversion as much as possible. It's easy to introduce bugs that way.

Do you think this is simpler?
```
if (v2OnlyAuthenticate) {
  return v2IsValidated
}
else {
  return v2IsValidated | fallbackValidateSignatureV1(mAuthRequest, clientPublicKey);
}
```

bgong-mdsol: fallbackValidateSignatureV1(...) should be called when v2IsValidated is false. To avoid boolean inversion, we can change as below 
```
if (v2OnlyAuthenticate) {
  return v2IsValidated;
}
else if (!v2IsValidated) {
  return fallbackValidateSignatureV1(mAuthRequest, clientPublicKey);
}
return v2IsValidated;

```

bgong-mdsol: Good point. **fallbackValidateSignatureV1**(...) should be called when v2IsValidated is false and not  v2OnlyAuthenticate. how do you think if simplify as below?
```
if (v2OnlyAuthenticate || v2IsValidated) {
  return v2IsValidated
}
else {
  return fallbackValidateSignatureV1(mAuthRequest, clientPublicKey);
}
```

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
```diff
@@ -168,4 +174,28 @@ private boolean validateSignatureV2 (MAuthRequest mAuthRequest, PublicKey client
     }
   }
 
+  private boolean fallbackValidateSignatureV1(MAuthRequest mAuthRequest, PublicKey clientPublicKey) {
```
jatcwang: Is it possible to reuse the logic in `validateSignatureV1`?

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
jatcwang: I want the code to follow the actual business logic as close as possible.

This is what I think the logic (in english) is:

```
If (v2OnlyAuthenticate)
  return result of v2 authentication
else
  return the result of v2 authentication, if that fails then return the result of v1 authentication
```

The code follows the logic

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
bgong-mdsol: The method fallbackValidateSignatureV1(MAuthRequest mAuthRequest, PublicKey clientPublicKey) calls validateSignatureV1, but also do pre-post as required:
- pre `validateSignatureV1`:
construct MAuthRequest `mAuthRequestV1` based on the original `mAuthRequest` with V1 headers (getXmwsSignature, getXmwsTime);
- call `validateSignatureV1` with `mAuthRequestV1`;
- post `validateSignatureV1`: write a log message `Completed successful authentication attempt after fallback to V1` if V1 is validated

jatcwang: Yes you're right I misread the implementation.

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
bgong-mdsol: updated as suggested https://github.com/mdsol/mauth-jvm-clients/commit/cbc0eca0e562d0b7b87d895e09e114f752da3b2a

jatcwang: Thanks.

PR #108, opened 2020-02-26
Switch to Scala 2.13

austek:
Switch to Scala 2.13 and add cross build for 2.12


.travis.yml
```diff
@@ -8,25 +8,23 @@ env:
...
-      script: sbt scalafmtSbtCheck scalafmtCheckAll +test +mimaReportBinaryIssues
-    - stage: release
-      script: 
-        - .travis/deploy.sh
-        - sbt mauth-proxy/ecr:push
-      if: tag IS NOT present
-stages:
-  - test
-  - release
+script: sbt scalafmtSbtCheck scalafmtCheckAll test
```
dgeyer-mdsol: was mimaReportBinaryIssues removed on purpose?

dgeyer-mdsol:
Thanks for upgrading this, maybe it would be helpful to add a dev-config/hooks/pre-commit-hook-install.sh explanation in the CONTRIBUTING file

.travis.yml
austek: good spot, reverted the change

PR #116, opened 2020-03-18
Address deprecation warnings

austek:
* Add scalac flags for 2.12 and 2.13
* Address deprecation warnings, silence
* Use sbt-smartrelease


build.sbt
```diff
@@ -199,5 +198,6 @@ lazy val `mauth-jvm-clients` = (project in file("."))
     publishSettings,
     crossScalaVersions := Nil,
     publish / skip := false,
-    mimaPreviousArtifacts := Set.empty
+    mimaPreviousArtifacts := Set.empty,
```
dgeyer-mdsol: I think this one if from the mima plugin that was removed

austek: removed, now it's handled by `smartReleaseAggregateProjectSettings`

PR #121, opened 2020-04-02
MCC-603053 Use getRawPath to get the path of the URL when using apach…

abinesh-mdsol:
Use getRawPath to get the path of the URL when using apache http for signing the URL:

### DIFF
--- a/modules/mauth-signer-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientRequestSigner.java
+++ b/modules/mauth-signer-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientRequestSigner.java
@@ -82,7 +82,7 @@ public class HttpClientRequestSigner extends DefaultSigner {
       queryParams.append(param.getName()).append('=').append(param.getValue());
     }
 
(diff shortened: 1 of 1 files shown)


jfeltesse-mdsol:
cc @mdsol/architecture-enablement

modules/mauth-signer-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientRequestSigner.java
```diff
@@ -82,7 +82,7 @@ public void signRequest(HttpUriRequest request) throws MAuthSigningException {
       queryParams.append(param.getName()).append('=').append(param.getValue());
     }
 
-    Map<String, String> mauthHeaders = generateRequestHeaders(httpVerb, request.getURI().getPath(), body, queryParams.toString());
+    Map<String, String> mauthHeaders = generateRequestHeaders(httpVerb, request.getURI().getRawPath(), body, queryParams.toString());
```
austek, line 85: Could you add tests that addresses the use case please

austek:
Handled by #122

PR #123, opened 2020-05-18
[MCC-620147] : mauth-proxy changes to work with POST and PATCH

nsingh-mdsol:
This PR includes the changes:
  1. refactor mauth-proxy to accept POST and PATCH request. (it was failing with 
                 java.lang.UnsupportedOperationException: direct buffer at com.mdsol.mauth.proxy.MAuthHttpRequestSigner.signRequest(MAuthHttpRequestSigner.java:28)
  2. update the tests

### DIFF
--- a/modules/mauth-proxy/src/main/java/com/mdsol/mauth/proxy/MAuthHttpRequestSigner.java
+++ b/modules/mauth-proxy/src/main/java/com/mdsol/mauth/proxy/MAuthHttpRequestSigner.java
@@ -1,7 +1,6 @@
 package com.mdsol.mauth.proxy;
 
 import com.mdsol.mauth.Signer;
-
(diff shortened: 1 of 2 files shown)


jatcwang:
FYI you can now use https://github.com/mdsol/mauth-insomnia-plugin instead of mauth-proxy. The littleproxy library is no longer maintained which means we'll eventually have to remove it.

nsingh-mdsol:
Does this mean I should close this PR and start using `mauth-insomnia-plugin`?

jatcwang:
@nsingh-mdsol Can try and get this merged in case anyone else wants to use it, but if the insomnia plugin solves your problem then that's the better way forward

nsingh-mdsol:
@austek, addressed the comments. could you please re-look when you have a minute?

PR #124, opened 2020-05-29
[MCC-619935]mauth-protocol-test-suite

bgong-mdsol:
This PR added the tests of [mauth protocol test suite](https://github.com/mdsol/mauth-protocol-test-suite) for Signer and Authenticator.

@mdsol/architecture-enablement @austek

### DIFF
--- a/modules/mauth-authenticator-scala/src/test/scala/com/mdsol/mauth/RequestAuthenticatorProtocolSuiteSpec.scala
+++ b/modules/mauth-authenticator-scala/src/test/scala/com/mdsol/mauth/RequestAuthenticatorProtocolSuiteSpec.scala
@@ -1,0 +1,64 @@
+package com.mdsol.mauth
+
+import java.util.UUID
+
+import com.mdsol.mauth.test.utils.ProtocolTestSuiteHelper
+import com.mdsol.mauth.util.{EpochTimeProvider, MAuthKeysHelper}
+import com.mdsol.mauth.utils.ClientPublicKeyProvider
+import org.scalamock.scalatest.MockFactory
+import org.scalatest.BeforeAndAfterAll
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+class RequestAuthenticatorProtocolSuiteSpec extends AnyFlatSpec with BeforeAndAfterAll with Matchers with MockFactory {
+
+  val REQUEST_VALIDATION_TIMEOUT_SECONDS: Long = 300L
+  val mockClientPublicKeyProvider: ClientPublicKeyProvider = mock[ClientPublicKeyProvider]
+  val mockEpochTimeProvider: EpochTimeProvider = mock[EpochTimeProvider]
+  val authenticatorV2: RequestAuthenticator =
+    new RequestAuthenticator(mockClientPublicKeyProvider, REQUEST_VALIDATION_TIMEOUT_SECONDS, mockEpochTimeProvider, true)
+
+  it should "pass the tests in protocol test suite for MWSV2" in {
+    //noinspection ConvertibleToMethodValue
+    Option(ProtocolTestSuiteHelper.loadSigningConfig()) match {
+      case Some(signingConfig) =>
+        Option(ProtocolTestSuiteHelper.getTestCases) match {
+          case Some(testCases) =>
+            val publicKey = MAuthKeysHelper.getPublicKeyFromString(ProtocolTestSuiteHelper.getPublicKey)
+            for (caseName <- testCases) {
+              println("Test case: " + caseName)
+              val caseFile = caseName.concat("/").concat(caseName)
+              Option(ProtocolTestSuiteHelper.loadAuthenticationHeader(caseFile.concat(".authz"))) match {
+                case Some(authHeader) =>
+                  Option(ProtocolTestSuiteHelper.loadUnignedRequest(caseFile.concat(".req"))) match {
+                    case Some(unsignedRequest) =>
+                      val bodyInBytes = ProtocolTestSuiteHelper.getTestRequestBody(unsignedRequest, caseName)
+                      val mauthRequest = MAuthRequest.Builder.get
+                        .withAuthenticationHeaderValue(authHeader.getMccAuthentication)
+                        .withTimeHeaderValue(String.valueOf(authHeader.getMccTime))
+                        .withHttpMethod(unsignedRequest.getHttpVerb)
+                        .withResourcePath(unsignedRequest.getResourcePath)
+                        .withQueryParameters(unsignedRequest.getQueryString)
+                        .withMessagePayload(bodyInBytes)
+                        .build()
+
+                      (mockEpochTimeProvider.inSeconds _: () => Long).expects().returns(signingConfig.getRequestTime.toLong + 3)
+                      (mockClientPublicKeyProvider.getPublicKey _)
+                        .expects(UUID.fromString(signingConfig.getAppUuid))
+                        .returns(publicKey)
+                      authenticatorV2.authenticate(mauthRequest) shouldBe true
+                    case None =>
+                      println("No unsigned request available, skip")
+                  }
+                case None =>
+                  println("No Expected authentication available, skip")
+              }
+            }
+          case None =>
+            println("No test case available, done")
+        }
+      case None =>
+        println("No signing config available, done")
+    }
+  }
+}
--- a/modules/mauth-authenticator-scala/src/test/scala/com/mdsol/mauth/RequestAuthenticatorProtocolSuiteSpec.scala
+++ b/modules/mauth-authenticator-scala/src/test/scala/com/mdsol/mauth/RequestAuthenticatorProtocolSuiteSpec.scala
@@ -9,8 +9,9 @@ import org.scalamock.scalatest.MockFactory
 import org.scalatest.BeforeAndAfterAll
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
+import org.scalatest.prop.TableDrivenPropertyChecks
 
-class RequestAuthenticatorProtocolSuiteSpec extends AnyFlatSpec with BeforeAndAfterAll with Matchers with MockFactory {
+class RequestAuthenticatorProtocolSuiteSpec extends AnyFlatSpec with BeforeAndAfterAll with Matchers with TableDrivenPropertyChecks with MockFactory {
 
   val REQUEST_VALIDATION_TIMEOUT_SECONDS: Long = 300L
   val mockClientPublicKeyProvider: ClientPublicKeyProvider = mock[ClientPublicKeyProvider]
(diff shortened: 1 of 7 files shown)


jfeltesse-mdsol:
@mdsol/team_57 @jatcwang can you guys have a look? (got your names from the contributors list and it looks like you contributed recently)

modules/mauth-signer/src/test/scala/com/mdsol/mauth/SignerProtocolSuiteSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+    Option(ProtocolTestSuiteHelper.loadSigningConfig()) match {
+      case Some(signingConfig) =>
+        Option(ProtocolTestSuiteHelper.getTestCases) match {
+          case Some(testCases) =>
+            val uuid = UUID.fromString(signingConfig.getAppUuid)
+            val privateKey = getPrivateKeyFromString(signingConfig.getPrivateKey)
+            val mAuthSigner = new DefaultSigner(uuid, privateKey, mockEpochTimeProvider, true)
+
+            for (caseName <- testCases) {
+              println("Test case: " + caseName)
```
jatcwang, line 31: We probably want to make each of these an individual test case instead of all tests being run inside one single test so that one failing test does not kill every test after it. Can we split them up?

modules/mauth-test-utils/src/main/java/com/mdsol/mauth/test/utils/ProtocolTestSuiteHelper.java
```diff
@@ -0,0 +1,135 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class ProtocolTestSuiteHelper {
+
+  final static String TEST_SUITE_RELATIVE_PATH =
+      System.getenv("TEST_SUITE_RELATIVE_PATH") != null ? System.getenv("TEST_SUITE_RELATIVE_PATH")
+          : "../../../mauth-protocol-test-suite";
+  final static String MWSV2_TEST_CASE_PATH = TEST_SUITE_RELATIVE_PATH + "/protocols/MWSV2/";
```
jatcwang, line 23: Let's use git submodules for this otherwise your test failures are unreproducible.

modules/mauth-signer/src/test/scala/com/mdsol/mauth/SignerProtocolSuiteSpec.scala
bgong-mdsol: It is good suggestion. I am not sure how to split them up in scala, any suggestion?

modules/mauth-test-utils/src/main/java/com/mdsol/mauth/test/utils/ProtocolTestSuiteHelper.java
bgong-mdsol: Yes, once `mauth-protocol-test-suite` goes to public, will utilize git submodule

jfeltesse-mdsol: Let's have a follow up story once the suite is public and 🙈 for now.

jatcwang: Why does it need to be public for git submodules to work?

jfeltesse-mdsol: because this repo is public. it would work _for us_ but for anyone outside it wouldn't (not that there's anyone I guess but who knows?)

jatcwang: Oh duh yeah ofc I forgot this repo is public too :sweat_smile:

I'd rather wait for that repo to go public and then make sure this repo's CI test against it. Currently it just succeeds if no test cases are found so I don't think there's much point merge this?

modules/mauth-test-utils/src/main/java/com/mdsol/mauth/test/utils/ProtocolTestSuiteHelper.java
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  public static byte[] loadTestData(String filePath) {
+    filePath = MWSV2_TEST_CASE_PATH + filePath;
+    byte[] bytes = new byte[0];
+    try {
+      bytes = Files.readAllBytes(normalizePath(filePath));
+    } catch (IOException ex) {
+      System.out.println("Unable to load: " + filePath);
+    }
```
jatcwang, line 66: This needs to fail with an exception, not recover with an empty body......

modules/mauth-signer/src/test/scala/com/mdsol/mauth/SignerProtocolSuiteSpec.scala
jatcwang: Here's the basic idea:

```
testCases.foreach { testCase =>
  testCase.description in {
    // run the tests
  }
}
```

Also, if we fail to load something it should abort(fail) the whole test suite, not succeed. Throwing exceptions is fine in this case. Something like 

```
class RequestAuthenticatorProtocolSuiteSpec extends .. {
  val signingConfig = Option(ProtocolTestSuiteHelper.loadSigningConfig()).getOrElse(throw new Exception("Cannot load signing config"))
  val testCases = Option(ProtocolTestSuiteHelper.getTestCases).getOrElse(throw new Exception("Unable to load test cases"))
  // .... etc etc
}

```

modules/mauth-test-utils/src/main/java/com/mdsol/mauth/test/utils/ProtocolTestSuiteHelper.java
bgong-mdsol: It is a little bit tricky here, this method is used to read file for all, currently some test cases may not include file for signer or authentication([see example](https://github.com/mdsol/mauth-protocol-test-suite/tree/develop/protocols/MWSV2/authentication-only-after-decoding)), so we just test Signer and/or Authentication if the related files are existing and can be loaded properly.

modules/mauth-signer/src/test/scala/com/mdsol/mauth/SignerProtocolSuiteSpec.scala
bgong-mdsol: refactor as suggested in https://github.com/mdsol/mauth-jvm-clients/pull/124/commits/a91f9b5eca383d2e1967deb7470ae8810ba7d76d
At current stage, we only do test if required files are loaded properly, will skip test if the expected files are not available.

modules/mauth-test-utils/src/main/java/com/mdsol/mauth/test/utils/ProtocolTestSuiteHelper.java
jatcwang: Can we check that it exists before trying to load it, and not swallow the exception?

Or maybe we can have a clear separation of test type detection vs actual execution?
For example,

```
sealed trait TestCase

final case class TestWithStringBody(
  name: String,
  body: String,
  ...
) extends TestCase

final case class TestWithBinaryBody(
  name: String,
  body: Array[Byte]
  ...
)
```

That way we decouple the loading of the test cases from execution, and we're not swallowing exceptions which makes diagnosing test failures difficult.

I've also raised an issue in the test-suite repo about making test type detection easier
https://github.com/mdsol/mauth-protocol-test-suite/issues/4

modules/mauth-signer/src/test/scala/com/mdsol/mauth/SignerProtocolSuiteSpec.scala
```diff
@@ -0,0 +1,82 @@
...
+import com.mdsol.mauth.test.utils.ProtocolTestSuiteHelper
+import com.mdsol.mauth.util.MAuthKeysHelper.getPrivateKeyFromString
+import com.mdsol.mauth.util.{EpochTimeProvider, MAuthSignatureHelper}
+import org.bouncycastle.jce.provider.BouncyCastleProvider
+import org.scalamock.scalatest.MockFactory
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.prop.TableDrivenPropertyChecks
+
+import scala.collection.JavaConverters._
```
jatcwang, line 15: This is deprecated - can you pull in https://github.com/scala/scala-library-compat which will provide `import scala.collection.JavaConverters._` for both scala 2.12 and 2.13?

modules/mauth-authenticator-scala/src/test/scala/com/mdsol/mauth/RequestAuthenticatorProtocolSuiteSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+import java.util.UUID
+
+import com.mdsol.mauth.test.utils.ProtocolTestSuiteHelper
+import com.mdsol.mauth.util.{EpochTimeProvider, MAuthKeysHelper}
+import com.mdsol.mauth.utils.ClientPublicKeyProvider
+import org.scalamock.scalatest.MockFactory
+import org.scalatest.BeforeAndAfterAll
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.prop.TableDrivenPropertyChecks
```
jatcwang, line 12: Don't think you need this?

modules/mauth-signer/src/test/scala/com/mdsol/mauth/SignerProtocolSuiteSpec.scala
```diff
@@ -0,0 +1,82 @@
...
+import java.util.UUID
+
+import com.mdsol.mauth.test.utils.ProtocolTestSuiteHelper
+import com.mdsol.mauth.util.MAuthKeysHelper.getPrivateKeyFromString
+import com.mdsol.mauth.util.{EpochTimeProvider, MAuthSignatureHelper}
+import org.bouncycastle.jce.provider.BouncyCastleProvider
+import org.scalamock.scalatest.MockFactory
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.prop.TableDrivenPropertyChecks
```
jatcwang, line 13: Not needed?

modules/mauth-test-utils/src/main/java/com/mdsol/mauth/test/utils/ProtocolTestSuiteHelper.java
jfeltesse-mdsol: Would it be ok to merge as-is for now and have a follow-up story next sprint?

bgong-mdsol: @jatcwang thanks a lot, will do as suggested. I am going to implement "factory", so it will be reused easily for both core signing/authentication and httpclient integration

jatcwang: @bgong-mdsol If you do it in Scala you can leverage `sealed trait` so you can get exhaustive matching in the tests

```
testSpecifications.foreach { testSpec =>
  testSpec match {
    case BinaryBodyTest(...) => 
    case AuthenticationOnlyTest(...) =>
    // if you have a new test case you'll get compile time warning that you're not handling it 
 }
}
```

jatcwang:
Replaced by #133 I think

PR #128, opened 2020-06-24
[MCC-621192]Make the signing protocol versions configurable

bgong-mdsol:
This PR includes the changes:
1. replace `V2_ONLY_SIGN_REQUESTS` option with `MAUTH_SIGN_VERSIONS` option;
2. change the default to `v2` only for the signing protocol version;

**Notes:**
Since not all apps are updated to the latest, please use v1 when you experience authentication failures.
Ruby: https://github.com/mdsol/mauth-client-ruby/pull/40
Python: https://github.com/mdsol/mauth-client-python/pull/19

### DIFF
--- a/modules/mauth-proxy/src/main/java/com/mdsol/mauth/proxy/ProxyConfig.java
+++ b/modules/mauth-proxy/src/main/java/com/mdsol/mauth/proxy/ProxyConfig.java
@@ -1,21 +1,27 @@
 package com.mdsol.mauth.proxy;
 
+import com.mdsol.mauth.MAuthVersion;
+
 import com.typesafe.config.Config;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import scala.runtime.Static;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import java.util.UUID;
 
 public class ProxyConfig {
   private static final Logger logger = LoggerFactory.getLogger(ProxyConfig.class);
 
-  public static final String V2_ONLY_SIGN_REQUESTS = "mauth.v2_only_sign_requests";
-  public static final String V2_ONLY_AUTHTICATE = "mauth.v2_only_authenticate";
+  public static final String SIGN_VERSIONS = "mauth.sign_versions";
+  public static final String V2_ONLY_AUTHENTICATE = "mauth.v2_only_authenticate";
   private final int proxyPort;
   private final int bufferSizeInByes;
   private String privateKey;
   private final UUID appUuid;
-  private boolean v2OnlySignRequests = false;
+  private List<MAuthVersion> signVersions;
   private boolean v2OnlyAuthenticate = false;
 
   public ProxyConfig(Config config) {
@@ -24,18 +30,18 @@ public class ProxyConfig {
       config.getInt("proxy.buffer_size_in_bytes"),
       UUID.fromString(config.getString("app.uuid")),
       config.getString("app.private_key"),
-      config.getBoolean(V2_ONLY_SIGN_REQUESTS),
-      config.getBoolean(V2_ONLY_AUTHTICATE)
+      config.getString(SIGN_VERSIONS),
+      config.getBoolean(V2_ONLY_AUTHENTICATE)
     );
   }
 
   public ProxyConfig(int proxyPort, int bufferSizeInByes, UUID appUuid, String privateKey,
-      boolean v2OnlySignRequests, boolean v2OnlyAuthenticate) {
+                     String signVersions, boolean v2OnlyAuthenticate) {
     this.proxyPort = proxyPort;
     this.bufferSizeInByes = bufferSizeInByes;
     this.appUuid = appUuid;
     this.privateKey = privateKey;
-    this.v2OnlySignRequests = v2OnlySignRequests;
+    this.signVersions = getSignVersions(signVersions);
     this.v2OnlyAuthenticate = v2OnlyAuthenticate;
   }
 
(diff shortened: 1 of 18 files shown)


modules/mauth-proxy/src/main/resources/reference.conf
```diff
@@ -11,6 +11,7 @@ app {
 }
 
 mauth {
-  v2_only_sign_requests: false
+  mauth_sign_versions: "v2" # default value
+  mauth_sign_versions: ${?MAUTH_SIGN_VERSIONS}
```
ykitamura-mdsol, line 15: IMHO, `mauth.mauth_sign_versions` sounds redundant.. 😅 

How about
```
sign_versions: ${?MAUTH_SIGN_VERSIONS}
```
here?

bgong-mdsol: reworded as suggested

modules/mauth-proxy/src/main/java/com/mdsol/mauth/proxy/ProxyConfig.java
```diff
@@ -54,8 +59,28 @@ public String getPrivateKey() {
   public UUID getAppUuid() {
     return appUuid;
   }
-  public boolean isV2OnlySignRequests() {
-    return v2OnlySignRequests;
+
+  public List<MAuthVersion> getSignVersions() { return signVersions; }
+
+  public List<MAuthVersion> getSignVersions(String signVersionsStr) {
```
jatcwang, line 65: this should be static?

modules/mauth-proxy/src/main/java/com/mdsol/mauth/proxy/ProxyConfig.java
```diff
@@ -54,8 +59,28 @@ public String getPrivateKey() {
...
+    List<MAuthVersion> signVersions = new ArrayList();
+    List<String> versionList = Arrays.asList(signVersionsStr.toUpperCase().split(","));
+    versionList.forEach(e -> {
+      switch (e.toLowerCase()) {
+        case "v1":
+          signVersions.add(MAuthVersion.MWS);
+          break;
+        case "v2":
+          signVersions.add(MAuthVersion.MWSV2);
+          break;
```
jatcwang, line 75: Can we log a warning if an unrecognized string is passed in?

modules/mauth-proxy/src/main/java/com/mdsol/mauth/proxy/ProxyConfig.java
```diff
@@ -54,8 +59,28 @@ public String getPrivateKey() {
...
-  public boolean isV2OnlySignRequests() {
-    return v2OnlySignRequests;
+
+  public List<MAuthVersion> getSignVersions() { return signVersions; }
+
+  public List<MAuthVersion> getSignVersions(String signVersionsStr) {
+    List<MAuthVersion> signVersions = new ArrayList();
+    List<String> versionList = Arrays.asList(signVersionsStr.toUpperCase().split(","));
+    versionList.forEach(e -> {
+      switch (e.toLowerCase()) {
```
jatcwang, line 69: Can we trim the input as well? it's common that we end up with an accidental space in the front or end and doing this can help avoid debugging sessions where you doubt your sanity :)

modules/mauth-signer/src/main/java/com/mdsol/mauth/SignerConfiguration.java
```diff
@@ -2,52 +2,81 @@
...
+        config.hasPath(MAUTH_SIGN_VERSIONS) ? config.getString(MAUTH_SIGN_VERSIONS) : "");
   }
 
   public SignerConfiguration(UUID appUUID, String privateKey) {
+    this(appUUID, privateKey, DEFAULT_SIGN_VERSION);
+  }
+
+  public SignerConfiguration(UUID appUUID, String privateKey, boolean v2OnlySignRequests) {
+    this(appUUID, privateKey, v2OnlySignRequests ? DEFAULT_SIGN_VERSION : ALL_SIGN_VERSIONS);
+  }
```
jatcwang, line 36: worth deprecating this?

modules/mauth-proxy/src/main/java/com/mdsol/mauth/proxy/ProxyConfig.java
jatcwang: All these comments apply to the DefaultSIgner too

bgong-mdsol: updated as suggested

modules/mauth-proxy/src/main/java/com/mdsol/mauth/proxy/ProxyConfig.java
bgong-mdsol: updated as suggeted

modules/mauth-signer/src/main/java/com/mdsol/mauth/SignerConfiguration.java
bgong-mdsol: I think it is better to remove this

modules/mauth-proxy/src/main/java/com/mdsol/mauth/proxy/ProxyConfig.java
bgong-mdsol: updated as suggested

ykitamura-mdsol:
FYI @jcarres-mdsol @cmcinnes-mdsol

modules/mauth-signer/src/main/java/com/mdsol/mauth/SignerConfiguration.java
```diff
@@ -1,53 +1,91 @@
...
-  public boolean isV2OnlySignRequests() {
-    return v2OnlySignRequests;
+  public List<MAuthVersion> getSignVersions() {
+    return signVersions;
+  }
+
+  static public List<MAuthVersion> getSignVersions(String signVersionsStr) {
+    List<MAuthVersion> signVersions = new ArrayList();
+    List<String> unrecognizedVersions = new ArrayList();
+    List<String> versionList = Arrays.asList(signVersionsStr.trim().toLowerCase().split(","));
```
jatcwang, line 67: this trim is unnecessary if you're already trimming every individual string after splitting

bgong-mdsol: removed in https://github.com/mdsol/mauth-jvm-clients/pull/128/commits/dc293690e580e527380a15b19f4cb9e3855d4d48

cmcinnes-mdsol:
hey @bgong-mdsol looking at the description again i see that we're defaulting to sign with only v2. I'd actually be more comfortable defaulting to signing with only V1 so that we can then turn on V2 later cc @ykitamura-mdsol

ykitamura-mdsol:
@cmcinnes-mdsol 

Here is the reason behind defaulting to sign with only v2:
https://github.com/mdsol/mauth-client-python/pull/19#discussion_r427633929

Unlike ruby, only a handful apps are using the java (and python) client so I think this is still manageable.

cmcinnes-mdsol:
Ok @ykitamura-mdsol, my concern was mainly for requests from java apps to ruby apps but if we observe problems we have the config options to handle it. I'm ok as is then 👍

PR #132, opened 2020-11-23
[MCC-693298]Accept request payload as InputStream for Java

bgong-mdsol:
This PR enhanced mauth-signer and mauth-authenticator to accept the request payload as InputStream.

Note: If payload is stream, 
- mauth-signer generates the signature for v2 only if v2 is required, or the signature for v1 if v1 only is required; 
- mauth-authenticator validates for v2 only if request headers including v2 signature, or validates for v1 if request headers including v1 signature only

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Accept request payload as java.io.InputStream for Java
 
 ## [5.0.2] - 2020-11-05
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
-- Accept request payload as java.io.InputStream for Java
+- Accept request payload as java.io.InputStream for Java. Since InputStream in general can only be consumed once, here are limitations of using stream payload:
+  - mauth-signer generates the signature for v2 only even the both v1 and v2 are required.
+  - mauth-authenticator doesn't support "Fall back to V1 authentication when V2 authentication fails".
 
 ## [5.0.2] - 2020-11-05
 ### Changed
--- a/modules/mauth-authenticator-scala/src/main/scala/com/mdsol/mauth/scaladsl/RequestAuthenticator.scala
+++ b/modules/mauth-authenticator-scala/src/main/scala/com/mdsol/mauth/scaladsl/RequestAuthenticator.scala
@@ -58,6 +58,8 @@ class RequestAuthenticator(publicKeyProvider: ClientPublicKeyProvider, epochTime
                   v2IsValidated
                 else if (v2IsValidated)
                   v2IsValidated
+                else if (mAuthRequest.getInputStream != null)
+                  v2IsValidated
                 else
                   fallbackValidateSignatureV1(mAuthRequest, clientPublicKey)
             }
@@ -86,14 +88,7 @@ class RequestAuthenticator(publicKeyProvider: ClientPublicKeyProvider, epochTime
     val decryptedSignature = MAuthSignatureHelper.decryptSignature(clientPublicKey, mAuthRequest.getRequestSignature)
     // Recreate the plain text signature, based on the incoming request parameters, and hash it.
     try {
-      val unencryptedRequestBytes = MAuthSignatureHelper.generateUnencryptedSignature(
-        mAuthRequest.getAppUUID,
-        mAuthRequest.getHttpMethod,
-        mAuthRequest.getResourcePath,
-        mAuthRequest.getMessagePayload,
-        String.valueOf(mAuthRequest.getRequestTime)
-      )
-      val messageDigest_bytes = MAuthSignatureHelper.getHexEncodedDigestedString(unencryptedRequestBytes).getBytes(StandardCharsets.UTF_8)
+      val messageDigest_bytes = MAuthSignatureHelper.generateDigestedMessageV1(mAuthRequest).getBytes(StandardCharsets.UTF_8)
 
       // Compare the decrypted signature and the recreated signature hashes.
       // If both match, the request was signed by the requesting application and is valid.
@@ -110,14 +105,7 @@ class RequestAuthenticator(publicKeyProvider: ClientPublicKeyProvider, epochTime
   private def validateSignatureV2(mAuthRequest: MAuthRequest, clientPublicKey: PublicKey): Boolean = {
     logAuthenticationRequest(mAuthRequest)
     // Recreate the plain text signature, based on the incoming request parameters, and hash it.
-    val unencryptedRequestString = MAuthSignatureHelper.generateStringToSignV2(
-      mAuthRequest.getAppUUID,
-      mAuthRequest.getHttpMethod,
-      mAuthRequest.getResourcePath,
-      mAuthRequest.getQueryParameters,
-      mAuthRequest.getMessagePayload,
-      String.valueOf(mAuthRequest.getRequestTime)
-    )
+    val unencryptedRequestString = MAuthSignatureHelper.generateStringToSignV2(mAuthRequest)
 
     // Compare the decrypted signature and the recreated signature hashes.
     try MAuthSignatureHelper.verifyRSA(unencryptedRequestString, mAuthRequest.getRequestSignature, clientPublicKey)
@@ -141,6 +129,7 @@ class RequestAuthenticator(publicKeyProvider: ClientPublicKeyProvider, epochTime
         mAuthRequest.getResourcePath,
         mAuthRequest.getQueryParameters
       )
+      mAuthRequestV1.setInputStream(mAuthRequest.getInputStream)
       isValidated = validateSignatureV1(mAuthRequestV1, clientPublicKey)
       if (isValidated) {
         logger.warn("Completed successful authentication attempt after fallback to V1")
--- a/modules/mauth-authenticator-scala/src/main/scala/com/mdsol/mauth/scaladsl/RequestAuthenticator.scala
+++ b/modules/mauth-authenticator-scala/src/main/scala/com/mdsol/mauth/scaladsl/RequestAuthenticator.scala
@@ -58,8 +58,6 @@ class RequestAuthenticator(publicKeyProvider: ClientPublicKeyProvider, epochTime
                   v2IsValidated
                 else if (v2IsValidated)
                   v2IsValidated
-                else if (mAuthRequest.getInputStream != null)
-                  v2IsValidated
                 else
                   fallbackValidateSignatureV1(mAuthRequest, clientPublicKey)
             }
@@ -129,7 +127,6 @@ class RequestAuthenticator(publicKeyProvider: ClientPublicKeyProvider, epochTime
         mAuthRequest.getResourcePath,
         mAuthRequest.getQueryParameters
       )
-      mAuthRequestV1.setInputStream(mAuthRequest.getInputStream)
       isValidated = validateSignatureV1(mAuthRequestV1, clientPublicKey)
       if (isValidated) {
         logger.warn("Completed successful authentication attempt after fallback to V1")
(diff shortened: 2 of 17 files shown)


austek (approved):

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
```diff
@@ -191,6 +180,7 @@ private boolean fallbackValidateSignatureV1(MAuthRequest mAuthRequest, PublicKey
           mAuthRequest.getResourcePath(),
           mAuthRequest.getQueryParameters()
       );
+      mAuthRequestV1.setInputStream(mAuthRequest.getInputStream());
```
jatcwang, line 183: if `fallbackValidateSignatureV1` isn't supposed to be called with an InputStream body, we should thrown an exception because it indicates a bug in MAuth.

modules/mauth-authenticator-scala/src/main/scala/com/mdsol/mauth/scaladsl/RequestAuthenticator.scala
```diff
@@ -141,6 +129,7 @@ class RequestAuthenticator(publicKeyProvider: ClientPublicKeyProvider, epochTime
         mAuthRequest.getResourcePath,
         mAuthRequest.getQueryParameters
       )
+      mAuthRequestV1.setInputStream(mAuthRequest.getInputStream)
```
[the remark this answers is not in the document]
bgong-mdsol: reword in https://github.com/mdsol/mauth-jvm-clients/pull/132/commits/3c0b8578c2ea43e3c379b8d895c7d551edda12a8

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
```diff
@@ -150,6 +152,14 @@ public MAuthVersion getMauthVersion() {
     return mauthVersion;
   }
 
+  public void setInputStream(InputStream inputStream) {
+    this.inputStream = inputStream;
+  }
```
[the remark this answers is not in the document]
bgong-mdsol: updated in https://github.com/mdsol/mauth-jvm-clients/pull/132/commits/3c0b8578c2ea43e3c379b8d895c7d551edda12a8

modules/mauth-authenticator/src/main/java/com/mdsol/mauth/RequestAuthenticator.java
bgong-mdsol: updated in https://github.com/mdsol/mauth-jvm-clients/pull/132/commits/3c0b8578c2ea43e3c379b8d895c7d551edda12a8

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
```diff
@@ -95,11 +130,19 @@ public MAuthRequest(String authenticationHeaderValue, byte[] messagePayload, Str
...
     this.httpMethod = httpMethod;
     this.requestTime = requestTime;
     this.resourcePath = resourcePath;
     this.queryParameters = queryParameters;
     this.mauthVersion = MAuthHeadersHelper.getMauthVersion(authenticationHeaderValue);
+
+    // we always use inputStream for Auth,
+    // wrap byte array in ByteArrayInputStream if bodyInputStream doesn't set
+    if (bodyInputStream == null) {
+      if (messagePayload == null) messagePayload = new byte[0];
```
jatcwang, line 143: I think not using a superclass like I suggested before is fine, but we need to make a clear distinction between:

* full message payload is not available (only stream)
* user specified full message payload (whether it's empty or not)

The first point is useful because it means you can actually abort V1 fallback auth with a meaningful message.

So I think what we need to do:

* In this primary constructor, throw an illegal argument exception if both messagePayload and bodyInputStream is not null
* In the constructor which takes a `byte[]` body, call the primary constructor with messagePayload = `new byte[0]` if the user passed in null as the body bytes
* V1 fallback validation should check that messagePayload == null, and if so just abort validation with a log message `"V1 authentication fallback is not available because the full request body is not available in memory"`

bgong-mdsol: Updated as suggested in https://github.com/mdsol/mauth-jvm-clients/pull/132/commits/54300d833460aad371e127fc8d8eeea8d10f69d9

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
```diff
@@ -139,9 +139,8 @@ private MAuthRequest(String authenticationHeaderValue, byte[] messagePayload, In
 
     // we always use inputStream for Auth,
     // wrap byte array in ByteArrayInputStream if bodyInputStream doesn't set
```
jatcwang, line 141:
```suggestion
    // wrap byte array in ByteArrayInputStream if bodyInputStream wasn't provided explicitly
```

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
```diff
@@ -139,9 +139,8 @@ private MAuthRequest(String authenticationHeaderValue, byte[] messagePayload, In
...
-    if (bodyInputStream == null) {
-      if (messagePayload == null) messagePayload = new byte[0];
-      this.bodyInputStream = new ByteArrayInputStream(messagePayload);
+    if (bodyInputStream == null && messagePayload != null) {
+       this.bodyInputStream = new ByteArrayInputStream(messagePayload);
     }
```
jatcwang, line 144: throw an exception in the else clause. Something is really wrong if the `if` branch did not match

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
```diff
@@ -77,6 +77,9 @@ public MAuthRequest(String authenticationHeaderValue, byte[] messagePayload, Str
   public MAuthRequest(String authenticationHeaderValue, byte[] messagePayload, String httpMethod,
       String timeHeaderValue, String resourcePath, String queryParameters) {
     this(authenticationHeaderValue, messagePayload, null, httpMethod, timeHeaderValue, resourcePath, queryParameters);
+    if (messagePayload == null) {
+      this.messagePayload = new byte[] {};
+    }
```
jatcwang, line 82: unnecessary if we put this logic in the primary constructor

bgong-mdsol: fixed https://github.com/mdsol/mauth-jvm-clients/pull/132/commits/16a30f1c09fed3a1590ed3b3e8cd106a367ce027

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
bgong-mdsol: updated in https://github.com/mdsol/mauth-jvm-clients/pull/132/commits/16a30f1c09fed3a1590ed3b3e8cd106a367ce027

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Accept request payload as java.io.InputStream for Java
```
jatcwang, line 9: Can we add some note about its limitations and what's not supported?

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
```diff
@@ -150,6 +195,14 @@ public MAuthVersion getMauthVersion() {
     return mauthVersion;
   }
 
+  public void setBodyInputStream(InputStream bodyInputStream) {
+    this.bodyInputStream = bodyInputStream;
+  }
```
jatcwang, line 200: Please remove this as it is no longer used and is problematic because it bypasses our constructor validation

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
```diff
@@ -85,21 +119,29 @@ public MAuthRequest(String authenticationHeaderValue, byte[] messagePayload, Str
...
-    this.messagePayload = messagePayload;
     this.httpMethod = httpMethod;
     this.requestTime = requestTime;
     this.resourcePath = resourcePath;
     this.queryParameters = queryParameters;
     this.mauthVersion = MAuthHeadersHelper.getMauthVersion(authenticationHeaderValue);
+
+    // we always use inputStream for Auth,
+    // wrap byte array in ByteArrayInputStream if bodyInputStream wasn't provided explicitly
+    validatePayload(messagePayload, bodyInputStream);
```
jatcwang, line 136: I don't think extracting this out helps readability. We want to inline it because it's only used in one place, and having the `if` conditions here makes it very clear what was checked 

Or put it another way: If you're a new reader reading this code, do you know what is checked by just reading this line and how it affects our assumption about the logic in the following lines?

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
```diff
@@ -239,8 +304,12 @@ public MAuthRequest build() {
         }
       }
 
+      if (messagePayload == null && bodyInputStream == null) {
+        messagePayload = new byte[] {};
+      }
```
jatcwang, line 309: no longer needed. Constructor already performs this

modules/mauth-common/src/test/scala/com/mdsol/mauth/MAuthRequestSpec.scala
```diff
@@ -282,4 +282,19 @@ class MAuthRequestSpec extends AnyFlatSpec with Matchers {
...
+        .get()
+        .withHttpMethod(CLIENT_REQUEST_METHOD)
+        .withAuthenticationHeaderValue(CLIENT_REQUEST_AUTHENTICATION_HEADER)
+        .withTimeHeaderValue(CLIENT_REQUEST_TIME_HEADER)
+        .withMessagePayload(CLIENT_REQUEST_PAYLOAD)
+        .withBodyInputStream(new java.io.ByteArrayInputStream(CLIENT_REQUEST_PAYLOAD))
+        .withResourcePath(CLIENT_REQUEST_PATH)
+        .build()
+    }
+    expectedException.getMessage shouldBe "Only one of bodyInputStream and messagePayload should be provided."
```
jatcwang, line 297: :+1:

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
bgong-mdsol: updated in https://github.com/mdsol/mauth-jvm-clients/pull/132/commits/ac9b72d9407033a365bbc428dfdd2315ead45826

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
bgong-mdsol: removed https://github.com/mdsol/mauth-jvm-clients/pull/132/commits/ac9b72d9407033a365bbc428dfdd2315ead45826

CHANGELOG.md
bgong-mdsol: added in https://github.com/mdsol/mauth-jvm-clients/pull/132/commits/ac9b72d9407033a365bbc428dfdd2315ead45826

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
bgong-mdsol: This case may occur when using Build(), move this part into Constructor. https://github.com/mdsol/mauth-jvm-clients/pull/132/commits/ac9b72d9407033a365bbc428dfdd2315ead45826

jatcwang: IMO the builder should be accessing the private constructor. But let's leave it there as it allows us to remove the code in the public constructor which takes a messagePayload

jatcwang (approved):

modules/mauth-common/src/main/java/com/mdsol/mauth/MAuthRequest.java
bgong-mdsol: This is the case that payload maybe not specified and no default value set either when using MAuthRequest.Builder.build() 
        MAuthRequest mAuthRequest = MAuthRequest.Builder.get()
		.withHttpMethod(method)
		.withResourcePath(requestPath)
		.withMauthHeaders(retrieveMauthHeaders(request))
		.withQueryParameters(queryParameters)
		.build();

PR #134, opened 2021-04-06
Update dependencies and remove little proxy

jatcwang:
Little proxy is unmaintained. There is https://github.com/mdsol/go-mauth-proxy which has faster startup and takes less memory


mayman (approved):

dgeyer-mdsol (approved):

project/build.properties
```diff
@@ -1,2 +1,2 @@
 # suppress inspection "UnusedProperty"
-sbt.version = 1.4.1
+sbt.version = 1.4.7
```
austek, line 2: 1.5.0?

austek (changes requested):
Needs documentation for users who needs to use proxy for maugh signing

project/build.properties
jatcwang: Can't go beyond 1.4.7 since travis CI's sbt launcher is still hardcoded to bintray (which is gone now)

austek (approved):

PR #135, opened 2021-06-26
[MCC-785068] Use Cache Control header as TTL

hballangan-mdsol:
- Honor Cache-Control max-age header as cache ttl
- Lazy load LoadingCache object to be able to set the ttl
- Add Caffeine library dependency for more efficient caching ([benchmarks](https://github.com/ben-manes/caffeine/wiki/Benchmarks))
- Add Cache-Control headers in FakeMauthServer response

### DIFF
--- a/modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
+++ b/modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
@@ -118,8 +118,7 @@ public class HttpClientPublicKeyProvider implements ClientPublicKeyProvider {
 
     public Optional<Long> getMaxAge(HttpResponse response) {
       return Optional.ofNullable(response.getFirstHeader(HttpHeaders.CACHE_CONTROL))
-        .map(Header::getElements)
-        .flatMap(elements -> Arrays.stream(elements)
+        .flatMap(header -> Arrays.stream(header.getElements())
           .filter(e -> e.getName().equalsIgnoreCase(MAX_AGE))
           .findFirst()
           .map(e -> Long.parseLong(e.getValue()))
--- a/modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
+++ b/modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
@@ -117,8 +117,8 @@ public class HttpClientPublicKeyProvider implements ClientPublicKeyProvider {
     }
 
     public Optional<Long> getMaxage(HttpResponse response) {
-      if (response.getFirstHeader(HttpHeaders.CACHE_CONTROL.toLowerCase()) != null) {
-        HeaderElement[] elements = response.getFirstHeader(HttpHeaders.CACHE_CONTROL.toLowerCase()).getElements();
+      if (response.getFirstHeader(HttpHeaders.CACHE_CONTROL) != null) {
+        HeaderElement[] elements = response.getFirstHeader(HttpHeaders.CACHE_CONTROL).getElements();
         return Arrays.stream(elements)
           .filter(e -> e.getName().equals(MAX_AGE))
           .findFirst()
--- a/modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
+++ b/modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
@@ -103,7 +103,7 @@ public class HttpClientPublicKeyProvider implements ClientPublicKeyProvider {
     public String handleResponse(HttpResponse response) throws IOException {
       if (response.getStatusLine().getStatusCode() == HttpStatus.SC_OK) {
         if (ttl == null) {
-          ttl = getMaxage(response).orElse(configuration.getTimeToLive());
+          ttl = getMaxAge(response).orElse(configuration.getTimeToLive());
         }
 
         HttpEntity entity = response.getEntity();
@@ -116,7 +116,7 @@ public class HttpClientPublicKeyProvider implements ClientPublicKeyProvider {
       }
     }
 
-    public Optional<Long> getMaxage(HttpResponse response) {
+    public Optional<Long> getMaxAge(HttpResponse response) {
       if (response.getFirstHeader(HttpHeaders.CACHE_CONTROL) != null) {
         HeaderElement[] elements = response.getFirstHeader(HttpHeaders.CACHE_CONTROL).getElements();
         return Arrays.stream(elements)
--- a/modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
+++ b/modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
@@ -117,14 +117,13 @@ public class HttpClientPublicKeyProvider implements ClientPublicKeyProvider {
     }
 
     public Optional<Long> getMaxAge(HttpResponse response) {
-      if (response.getFirstHeader(HttpHeaders.CACHE_CONTROL) != null) {
-        HeaderElement[] elements = response.getFirstHeader(HttpHeaders.CACHE_CONTROL).getElements();
-        return Arrays.stream(elements)
+      return Optional.ofNullable(response.getFirstHeader(HttpHeaders.CACHE_CONTROL))
+        .map(Header::getElements)
+        .flatMap(elements -> Arrays.stream(elements)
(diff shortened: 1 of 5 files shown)


modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
```diff
@@ -39,25 +37,20 @@
...
-            .expireAfterAccess(timeToLiveInSeconds, TimeUnit.SECONDS)
-            .build(new CacheLoader<UUID, PublicKey>() {
-              @Override
-              public PublicKey load(UUID appUUID) throws Exception {
-                return getPublicKeyFromEureka(appUUID);
-              }
-            });
+      Caffeine.newBuilder()
+        .expireAfterAccess(ttl, TimeUnit.SECONDS)
+        .build(this::getPublicKeyFromEureka);
```
jfeltesse-mdsol, line 53: This should probably be renamed `getPublicKeyFromMauth`, mauth clients shouldn't depend on eureka at all.

modules/mauth-test-utils/src/main/java/com/mdsol/mauth/test/utils/FakeMAuthServer.java
```diff
@@ -37,7 +38,7 @@ public static void resetMappings() {
 
   public static void return200() {
     WireMock.stubFor(WireMock.get(WireMock.urlPathEqualTo("/mauth/v1/security_tokens/" + EXISTING_CLIENT_APP_UUID.toString() + ".json"))
-        .willReturn(WireMock.aResponse().withStatus(200).withBody(mockedMauthTokenResponse())));
+        .willReturn(WireMock.aResponse().withStatus(200).withBody(mockedMauthTokenResponse()).withHeader(HttpHeaders.CACHE_CONTROL, "max-age=3600, private")));
```
jfeltesse-mdsol, line 41: please keep the lines within 120 chars

jfeltesse-mdsol (commented):
cc @mdsol/enablement-java

modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
hballangan-mdsol: This was on the old code but yes I do agree to change it. Already changed in my new commit.

bgong-mdsol:
Could you add the changes in CHANGELOG [Unreleased]?

hballangan-mdsol:
> Could you add the changes in CHANGELOG [Unreleased]?

hi @bgong-mdsol, already added the CHANGELOG

dgeyer-mdsol (approved):

bgong-mdsol (approved):
👍

modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
```diff
@@ -96,18 +96,35 @@ private String getRequestUrlPath(UUID appUUID) {
...
+      }
+    }
+
+    public Optional<Long> getMaxage(HttpResponse response) {
+      if (response.getFirstHeader(HttpHeaders.CACHE_CONTROL) != null) {
+        HeaderElement[] elements = response.getFirstHeader(HttpHeaders.CACHE_CONTROL).getElements();
+        return Arrays.stream(elements)
+          .filter(e -> e.getName().equals(MAX_AGE))
+          .findFirst()
+          .map(e -> Long.parseLong(e.getValue()));
```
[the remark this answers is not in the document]
hballangan-mdsol: Response cache header max-age value can only be a numeric value in seconds so there is no way this would fail unless the cache-headers are tampered with. So should we allow public key storage of response with tampered headers?

jfeltesse-mdsol:
by the way the version should be bumped and then the changelog adjusted accordingly. I don't think we have any feature or bugfix coming up so might as well cut a release with those changes.

modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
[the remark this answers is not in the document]
dgeyer-mdsol: Considering the tampered headers scenario makes sense to throw an exception. So the request to the service consuming this library will return 401.
Do you think there is a possibility of the Mauth service returning an invalid value here?, because in that case, the consequence will be all consumers of this library returning 401 for all requests, which could be avoided by using configuration.getTimeToLive()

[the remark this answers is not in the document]
hballangan-mdsol: Max-age can only be numeric, and it's impossible for mauth to have the wrong value since the Ruby framework is responsible for setting up the cache control headers (and making sure that the 'expires' value is numeric). Also there is a cucumber test in mauth to make sure that the value is valid.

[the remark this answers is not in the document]
jfeltesse-mdsol: Looks like we're covered then, merging EOD unless further objections.

jfeltesse-mdsol (changes requested):
do not merge: version bump and changelog update are still needed

jfeltesse-mdsol:
@hballangan-mdsol what's the rationale for that last commit about lowercase headers, have a link to share? AFAIK none of the platform services is http2 ready, and IIRC http headers are to be processed in a case insensitive way. Would that last commit make headers like `Cache-Control` not recognized or is the reverse true (e.g. the code was broken without lower casing)?

modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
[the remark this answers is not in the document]
dgeyer-mdsol: Looks good to me, thanks for checking

dgeyer-mdsol (approved):

hballangan-mdsol:
> insensitive

@jfeltesse-mdsol Ah sorry, I thought that push failed. I realized that HttpMessage.getFirstHeader uses equalsIgnoreCase to compare so there is no need to add lowercase. I will be remove the change.

modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
```diff
@@ -96,18 +96,35 @@ private String getRequestUrlPath(UUID appUUID) {
...
+        return mapper.readTree(responseAsString).findValue(PUBLIC_KEY_STR).asText();
       } else {
         throw new HttpClientPublicKeyProviderException("Invalid response code returned by server: "
-            + response.getStatusLine().getStatusCode());
+          + response.getStatusLine().getStatusCode());
+      }
+    }
+
+    public Optional<Long> getMaxAge(HttpResponse response) {
+      if (response.getFirstHeader(HttpHeaders.CACHE_CONTROL) != null) {
```
jatcwang, line 120: I think this whole method can be simplified to:
```
      return Optional.ofNullable(response.getFirstHeader(HttpHeaders.CACHE_CONTROL))
              .map(header -> Long.parseLong(header.getValue()));
```

hballangan-mdsol: Thanks for suggesting Optional.ofNullable in the first part! I included it in my commit to remove the if else statement, but aside from that,  I don't think the logic can be simplified further because I need to go through the cache-control header elements to get the max-age.

jatcwang: Yes sorry I think I misunderstood the intention. 

Reading this a bit closer and looking at the HTTP header spec, I think it is valid to have the values split across multiple HTTP headers:

```
Cache-Control: public
Cache-Control: max-age=10000
```
(which is semantically equivalent to `Cache-Control: public, max-age=10000`)

In which case `getFirstHeader` will only give us `public`. Perhaps we should switch to using response.getHeaders(HttpHeaders.CACHE_CONTROL).flatMap(..) to handle this possibility?

hballangan-mdsol: Oh wow I even didn't think that was even possible, so the exception is for headers with comma-separated values. I have resolved that scenario in my new commit.

jatcwang (approved):

jatcwang:
@jfeltesse-mdsol feel free to merge (I cannot merge since you have requested changes)

jfeltesse-mdsol (approved):

PR #141, opened 2022-07-29
Create http4s mAuth middleware

fserra-mdsol:


modules/mauth-authenticator-http4s/src/test/scala/com/mdsol/mauth/http4s/MAuthMiddlewareSuite.scala
```diff
@@ -0,0 +1,268 @@
...
+      .assert
+  }
+
+  test("reject if public key cannot be found") {
+    val localClient = new ClientPublicKeyProvider {
+      override def getPublicKey(appUUID: UUID): Future[Option[PublicKey]] =
+        if (appUUID == appUuid) {
+          Future(none)
+        } else Future.failed(new Throwable("Wrong app UUID"))
+      override def getPublicKeyIO(appUUID: UUID): IO[Option[PublicKey]] = ???
```
jatcwang, line 147: Can you remove this from the public interface while you're at it? :) I don't think it's suppose to be public

fserra-mdsol: Yep, on it!

jatcwang (approved):

jatcwang (approved):

modules/mauth-authenticator-http4s/src/test/scala/com/mdsol/mauth/http4s/MAuthMiddlewareSuite.scala
tmilner: I think we did it on purpose so you dont need to use future in places like this. The future API is just calling this with an .unsafeToFuture.

tmilner (approved):

modules/mauth-authenticator-http4s/src/test/scala/com/mdsol/mauth/http4s/MAuthMiddlewareSuite.scala
tmilner: If its not used anywhere though its fine to remove it.

PR #142, opened 2022-08-03
[MCC-942718]Enhance MAuth Authenticator Apachehttp caching

bgong-mdsol:
This PR fixed the issue reported in https://mdsol.slack.com/archives/CCDANR533/p1656530149481259
 
@mdsol/architecture-enablement @jatcwang @johnduhart

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Enhanced the caching implementation of MAuth Authenticator Apachehttp
 
 ## [9.0.0] - 2022-07-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Changed
-- Enhanced the caching implementation of MAuth Authenticator Apachehttp
+- Fix caching to properly re-fetch from MAuth Service in MAuth Authenticator Apachehttp
 
 ## [9.0.0] - 2022-07-29
 
--- a/modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
+++ b/modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
@@ -44,6 +44,7 @@ public class HttpClientPublicKeyProvider implements ClientPublicKeyProvider {
     this.signer = signer;
     this.httpclient = HttpClients.createDefault();
     this.publicKeyResponseHandler = new PublicKeyResponseHandler();
+    setupCache();
   }
 
   private void setupCache() {
@@ -72,17 +73,7 @@ public class HttpClientPublicKeyProvider implements ClientPublicKeyProvider {
 
   @Override
   public PublicKey getPublicKey(UUID appUUID) {
-    try {
-      if (publicKeyCache == null) {
-        PublicKeyData data = getPublicKeyFromMauth(appUUID);
-        setupCache();
-        publicKeyCache.put(appUUID, data);
-      }
-      return publicKeyCache.get(appUUID).getPublicKey();
-    } catch (Exception e) {
-      logger.error("Public key retrieval error", e);
-      throw new HttpClientPublicKeyProviderException(e);
-    }
+    return publicKeyCache.get(appUUID).getPublicKey();
   }
 
   private String getRequestUrlPath(UUID appUUID) {
@@ -97,6 +88,7 @@ public class HttpClientPublicKeyProvider implements ClientPublicKeyProvider {
       }
       return httpclient.execute(httpGet, responseHandler);
     } catch (IOException ex) {
+      logger.error("Public key retrieval error", ex);
       throw new HttpClientPublicKeyProviderException(ex);
     }
   }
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Enhanced the caching implementation of MAuth Authenticator Apachehttp
```
johnduhart, line 9: Can you improve the wording here? "Enhanced" is vague, we're fixing a bug around caching.

modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
```diff
@@ -102,9 +105,7 @@ private class PublicKeyResponseHandler implements ResponseHandler<String> {
     @Override
     public String handleResponse(HttpResponse response) throws IOException {
       if (response.getStatusLine().getStatusCode() == HttpStatus.SC_OK) {
-        if (ttl == null) {
-          ttl = getMaxAge(response).orElse(configuration.getTimeToLive());
-        }
+        ttl = getMaxAge(response).orElse(configuration.getTimeToLive());
```
johnduhart, line 108: Do the additional writes to `ttl` have an effect if the cache was already setup?

CHANGELOG.md
bgong-mdsol: updated in https://github.com/mdsol/mauth-jvm-clients/pull/142/commits/dd0570b356fba0741430abc554e7265a8a7a554b

modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
bgong-mdsol: See https://github.com/mdsol/mauth-jvm-clients/pull/142/commits/dd0570b356fba0741430abc554e7265a8a7a554b

modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
```diff
@@ -102,9 +116,7 @@ private class PublicKeyResponseHandler implements ResponseHandler<String> {
...
```
jatcwang, line 119: I think we probably want to fix more things in this class. 

ttl should not be a field that we modify on every request sent - this class is going to be used in a multithreaded context so it's just not right..

Here's what I suggest we do:

* Make PublicKeyResponseHandler return a String (the pub key) and a Long (the age)
* Make the cache store both the pub key and the max age, you can then use [Expiry](https://github.com/ben-manes/caffeine/wiki/Eviction#time-based) to calculate the expiry time (since max age is now part of the cache value)
* Remove the whole mutex thing. It's going to lead to massive performance problems as only one caller can call `getPublicKey` at a time...

bgong-mdsol: @jatcwang thanks a lot. Updated the cache as suggested in https://github.com/mdsol/mauth-jvm-clients/pull/142/commits/4fded8e97d7b44f386b8f81b1bf50eda53fbe254

jatcwang (approved):

modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
```diff
@@ -47,32 +47,38 @@ public HttpClientPublicKeyProvider(AuthenticatorConfiguration configuration, Sig
   }
 
   private void setupCache() {
-    publicKeyCache =
-      Caffeine.newBuilder()
-        .expireAfterAccess(ttl, TimeUnit.SECONDS)
+    publicKeyCache = Caffeine.newBuilder()
```
johnduhart, line 50: There's no longer a need to defer creating the cache until the first request, can you please move this to the constructor?

bgong-mdsol: updated in https://github.com/mdsol/mauth-jvm-clients/pull/142/commits/a94bc4936cacee46ca89bac115327abf5aac9f04

modules/mauth-authenticator-apachehttp/src/main/java/com/mdsol/mauth/apache/HttpClientPublicKeyProvider.java
```diff
@@ -35,48 +37,43 @@ public class HttpClientPublicKeyProvider implements ClientPublicKeyProvider {
   private final Signer signer;
   private final CloseableHttpClient httpclient;
   private final PublicKeyResponseHandler publicKeyResponseHandler;
-
-  private LoadingCache<UUID, PublicKey> publicKeyCache;
-  private Long ttl;
+  private LoadingCache<UUID, PublicKeyData> publicKeyCache;
```
johnduhart, line 40: This can be made a final now, it only needs to be set in the constructor.

bgong-mdsol: updated in https://github.com/mdsol/mauth-jvm-clients/pull/142/commits/907273382ec3d2c60482ca7168f88b6e391bd96b

johnduhart (approved):

PR #148, opened 2023-05-10
add http4s MauthPublicKeyProvider

mayman:
Add MauthPublicKeyProvider for Http4s


modules/mauth-signer-http4s/src/main/scala/com/mdsol/mauth/http4s/client/Implicits.scala
```diff
@@ -0,0 +1,46 @@
...
+      val headersWithoutContentType: Map[String, String] = removeContentTypeFromHeaders(signedRequest.req.headers)
+
+      val allHeaders: immutable.Seq[Header.Raw] = (headersWithoutContentType ++ signedRequest.mauthHeaders).toList
+        .map { case (name, value) =>
+          Header.Raw(CIString(name), value)
+        }
+
+      Request[F](
+        method = Method.fromString(signedRequest.req.httpMethod).getOrElse(Method.GET),
+        uri = Uri(path = Uri.Path.unsafeFromString(signedRequest.req.uri.toString)),
```
fserra-mdsol, line 28: i understand why you'd want to do this, but (especially) in library code we should avoid doing these 2 things: forcing a fallback value and using unsafe methods. The rationale is that debugging potential issues coming from this would be ridiculously difficult. 
I'd claim that this method's signature should be `def toHttp4sRequest[F[_]]: F[Request[F]]`, in order to delegate to the caller the decision on what to do for:
1. whether the method used from within the wrapped request instance is not derivable as canonically named by http4s
2. the URI can't be built based on the supplied String representation

modules/mauth-signer-http4s/src/main/scala/com/mdsol/mauth/http4s/client/Implicits.scala
```diff
@@ -0,0 +1,46 @@
...
+        body = fs2.Stream.emits(signedRequest.req.body),
+        headers = Headers(allHeaders)
+      ).withContentTypeOption(contentType)
+    }
+
+    private def extractContentTypeFromHeaders(requestHeaders: Map[String, String]): Option[`Content-Type`] =
+      requestHeaders
+        .get(headers.`Content-Type`.toString)
```
fserra-mdsol, line 36: I'd claim you shouldn't be using `headers.`Content-Type`` from `http4s` is the header you're trying to intercept is foreign to it

fserra-mdsol (changes requested):
Very nicely done! You're making me happy with this piece of work!

fserra-mdsol (changes requested):
Very nicely done! You're making me happy with this piece of work!

modules/mauth-authenticator-http4s/src/test/scala/com/mdsol/mauth/http4s/MauthPublicKeyProviderSuite.scala
```diff
@@ -0,0 +1,61 @@
...
+class MauthPublicKeyProviderSuite extends CatsEffectSuite {
+
+  private val MAUTH_PORT = PortFinder.findFreePort()
+  private val MAUTH_BASE_URL = s"http://localhost:$MAUTH_PORT"
+  private val MAUTH_URL_PATH = "/mauth/v1"
+  private val SECURITY_TOKENS_PATH = "/security_tokens/%s.json"
+  def executeRequest(uuid: String, response: IO[Response[IO]]): Kleisli[IO, Request[IO], Response[IO]] =
+    HttpRoutes
+      .of[IO] {
+        case GET -> _ / _ / _ / _ / "mauth" / "v1" / "security_tokens" / appId if appId == s"$uuid.json"                                  => response
```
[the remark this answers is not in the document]
mayman: This doesn't appear to work for some reason :(

modules/mauth-authenticator-http4s/src/main/scala/com/mdsol/mauth/http4s/RequestAuthenticator.scala
```diff
@@ -0,0 +1,155 @@
...
+    } else if (!validateMauthVersion(mAuthRequest, v2OnlyAuthenticate)) {
+      val message = "The service requires mAuth v2 authentication headers."
+      logger.error(message) *> Async[F].raiseError(new MAuthValidationException(message))
+    } else {
+      getPublicKey(mAuthRequest)
+    }
+  }
+
+  private def getPublicKey(mAuthRequest: MAuthRequest): F[Boolean] = {
+    publicKeyProvider.getPublicKey(mAuthRequest.getAppUUID).map {
```
fserra-mdsol, line 54: Looks like you need to `flatMap` here

modules/mauth-authenticator-http4s/src/main/scala/com/mdsol/mauth/http4s/RequestAuthenticator.scala
```diff
@@ -0,0 +1,155 @@
...
+      case None =>
+        logger.error("Public Key couldn't be retrieved")
+        false
```
fserra-mdsol, line 57: this wouldn't work - when you operate in some `F` context that gets evaluated by the cats effect runtime, all operations need to be chained. So typically this would look like
`logger.error(...) *> false.pure`

modules/mauth-authenticator-http4s/src/main/scala/com/mdsol/mauth/http4s/RequestAuthenticator.scala
```diff
@@ -0,0 +1,155 @@
...
+            else
+              fallbackValidateSignatureV1(mAuthRequest, clientPublicKey)
+        }
+    }
+  }
+
+}
+
+object RequestAuthenticator {
+  def apply[F[_]: Async](publicKeyProvider: ClientPublicKeyProvider[F], epochTimeProvider: EpochTimeProvider) =
```
fserra-mdsol, line 79: You can drop the constraint here too, consequently.

modules/mauth-authenticator-http4s/src/main/scala/com/mdsol/mauth/http4s/RequestAuthenticator.scala
```diff
@@ -0,0 +1,155 @@
...
+
+      // Compare the decrypted signature and the recreated signature hashes.
+      // If both match, the request was signed by the requesting application and is valid.
+      util.Arrays.equals(messageDigest_bytes, decryptedSignature)
+    } catch {
+      case ex: Exception =>
+        val message = "MAuth request validation failed for V1."
+        logger.error(ex)(message)
+        throw new MAuthValidationException(message, ex)
+    }
```
fserra-mdsol, line 106: Nope - you can definitely do better than try/catch and rethrow (we're in cats effect land, you can mark failed computations as failed `F`, so that the compiler has full control over the correctness of your program). Also, mind the chaining of the logger statements, otherwise they won't be executed

modules/mauth-authenticator-http4s/src/main/scala/com/mdsol/mauth/http4s/RequestAuthenticator.scala
```diff
@@ -0,0 +1,155 @@
...
+    isValidated
+  }
+
+  private def logAuthenticationRequest[F[_]](mAuthRequest: MAuthRequest)(implicit logger: SelfAwareStructuredLogger[F]): Unit = {
+    val msgFormat = "Mauth-client attempting to authenticate request from app with mauth app uuid %s using version %s.".format(
+      mAuthRequest.getAppUUID,
+      mAuthRequest.getMauthVersion.getValue
+    )
+    logger.info(msgFormat)
```
fserra-mdsol, line 153: Ouch, this would not even be executed :-(

modules/mauth-authenticator-http4s/src/test/scala/com/mdsol/mauth/http4s/MauthPublicKeyProviderSuite.scala
fserra-mdsol, line 1: Lovely!

modules/mauth-authenticator-http4s/src/test/scala/com/mdsol/mauth/http4s/RequestAuthenticatorSpec.scala
```diff
@@ -0,0 +1,151 @@
...
+  test("not authenticate a request after timeout period passed") {
+    val client: ClientPublicKeyProvider[IO] = _ => IO.none
+
+    val authenticator = RequestAuthenticator(client, mockEpochTimeProvider(CLIENT_UNICODE_X_MWS_TIME_HEADER_VALUE.toLong + 600))
+
+    authenticator.authenticate(getRequestWithUnicodeCharactersInBody).handleErrorWith {
+      case e: MAuthValidationException =>
+        IO.pure(assertEquals(e.getMessage, "MAuth request validation failed because of timeout 10 seconds"))
+      case _ => fail("should not be here")
+    }
```
fserra-mdsol, line 66: unfortunately this doesn't work as it would seem. The `assertEquals` inside the `IO` is never executed.
The correct way to prove that this works correctly is to call 
`authenticator.authenticate(getRequestWithUnicodeCharactersInBody).attempt` and deal with the disjoint union.
`assert` / `assertEqual` has to be the last instruction evaluated in the test scenario in order to get a valid test result (here instead it's nested within the `IO`, to clarify what I mean)

modules/mauth-authenticator-http4s/src/main/scala/com/mdsol/mauth/http4s/RequestAuthenticator.scala
```diff
@@ -0,0 +1,155 @@
...
+import com.mdsol.mauth.util.{EpochTimeProvider, MAuthSignatureHelper}
+import org.typelevel.log4cats.SelfAwareStructuredLogger
+import org.typelevel.log4cats.slf4j.Slf4jLogger
+
+import java.nio.charset.StandardCharsets
+import java.security.PublicKey
+import java.util
+import scala.concurrent.duration._
+
+class RequestAuthenticator[F[_]: Async](
```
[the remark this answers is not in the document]
mayman: I am not sure if I am doing something wrong, but when I drop it to ApplicativeThrow I can't flat map on it 😕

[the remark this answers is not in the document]
fserra-mdsol: Ohh, sorry, I didn't realise you were flatMapping over it - then you need `MonadThrow`! my bad

modules/mauth-signer-http4s/src/main/scala/com/mdsol/mauth/http4s/client/Implicits.scala
mayman: here we are just extracting the content type header, so I don't think it's doing anything foreign

modules/mauth-authenticator-http4s/src/test/scala/com/mdsol/mauth/http4s/RequestAuthenticatorSpec.scala
tmilner: Oh thats fun. It would work in out IOWordSpec scalatest suite :D

modules/mauth-authenticator-http4s/src/main/scala/com/mdsol/mauth/http4s/MauthPublicKeyProvider.scala
```diff
@@ -0,0 +1,80 @@
...
+import java.security.PublicKey
+import java.util.UUID
+import scala.concurrent.duration._
+import cats.implicits._
+import com.mdsol.mauth.http4s.MauthPublicKeyProvider.SecurityToken
+import io.circe.{Decoder, HCursor}
+import org.http4s.circe.jsonOf
+import org.typelevel.log4cats.Logger
+
+class MauthPublicKeyProvider[F[_]: Async: Logger](configuration: AuthenticatorConfiguration, signer: MAuthRequestSigner, val client: Client[F])(implicit
```
fserra-mdsol, line 25: this can be then 
```suggestion
class MauthPublicKeyProvider[F[_]: MonadThrow: Logger](configuration: AuthenticatorConfiguration, signer: MAuthRequestSigner, val client: Client[F])(implicit
```

modules/mauth-authenticator-http4s/src/main/scala/com/mdsol/mauth/http4s/MauthPublicKeyProvider.scala
```diff
@@ -0,0 +1,80 @@
...
+  val caffeineCache: Cache[F, String, Option[PublicKey]]
```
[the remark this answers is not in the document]
mayman: it just cache now, the name is missleading - will update ( I have separated the dependencies too)

modules/mauth-authenticator-http4s/src/main/scala/com/mdsol/mauth/http4s/RequestAuthenticator.scala
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+}
+
+object RequestAuthenticator {
+  def apply[F[_]: Async: Logger](publicKeyProvider: ClientPublicKeyProvider[F], epochTimeProvider: EpochTimeProvider) =
+    new RequestAuthenticator(publicKeyProvider, epochTimeProvider, v2OnlyAuthenticate = false)
+
+  // Check epoch time is not older than specified interval.
+  private def validateTime[F[_]: Async](requestTime: Long, epochTimeProvider: EpochTimeProvider)(requestValidationTimeout: Duration): F[Boolean] =
```
fserra-mdsol, line 82:
```suggestion
  private def validateTime[F[_]: Applicative](requestTime: Long, epochTimeProvider: EpochTimeProvider)(requestValidationTimeout: Duration): F[Boolean] =
```

fserra-mdsol: This can also be non-private, so that you can prove that it works correctly in a test

modules/mauth-authenticator-http4s/src/main/scala/com/mdsol/mauth/http4s/RequestAuthenticator.scala
```diff
@@ -0,0 +1,151 @@
...
+    ((epochTimeProvider.inSeconds - requestTime) < requestValidationTimeout.toSeconds).pure[F]
+
+  // Check V2 header if only V2 is required
+  private def validateMauthVersion[F[_]: Async](mAuthRequest: MAuthRequest, v2OnlyAuthenticate: Boolean): F[Boolean] =
```
fserra-mdsol, line 86:
```suggestion
  private def validateMauthVersion[F[_]: Applicative](mAuthRequest: MAuthRequest, v2OnlyAuthenticate: Boolean): F[Boolean] =
```

same with regards to non-private

modules/mauth-signer-http4s/src/main/scala/com/mdsol/mauth/http4s/client/Implicits.scala
tmilner: The header types though arent the http4s headers, they're just strings. I think its ok though as its just a constant you'd have to define again.

tmilner (approved):
Looks much better :)

modules/mauth-signer-http4s/src/main/scala/com/mdsol/mauth/http4s/client/Implicits.scala
fserra-mdsol: Sorry, I think my comment was really unclear. What I mean is that
1. `requestHeaders` is a Map of String representations of non-http4s headers
2. we're asking to get us the value of the `Content-Type` header based on the string representation of http4s
3. because of that you might end up with a mismatch.
Now, this is a minor thing, so please don't get blocked on this. Just wanted to clarify what I meant :-)

fserra-mdsol (approved):

modules/mauth-authenticator-http4s/src/test/scala/com/mdsol/mauth/http4s/RequestAuthenticatorSpec.scala
fserra-mdsol: Yeah, that's a quirk of MUnit. It's really easy to overlook that.

mayman: MUnit had a useful intercept that works a treat and it looks cleaner tbh

PR #156, opened 2023-07-17
Tech: Fix Cache to reduce requests.

tmilner:
Ensure we dont make additional requests when we get a burst of requests against an empty cache.

### DIFF
--- a/modules/mauth-authenticator-http4s/src/main/scala/com/mdsol/mauth/http4s/MauthPublicKeyProvider.scala
+++ b/modules/mauth-authenticator-http4s/src/main/scala/com/mdsol/mauth/http4s/MauthPublicKeyProvider.scala
@@ -1,7 +1,7 @@
 package com.mdsol.mauth.http4s
 
 import cats.ApplicativeThrow
-import cats.effect.{Async, Concurrent, Outcome, Sync}
+import cats.effect.{Async, Outcome, Sync}
 import com.mdsol.mauth.http4s.client.Implicits.NewSignedRequestOps
 import com.mdsol.mauth.models.UnsignedRequest
 import com.mdsol.mauth.scaladsl.utils.ClientPublicKeyProvider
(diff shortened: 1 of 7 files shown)


modules/mauth-authenticator-http4s/src/main/scala/com/mdsol/mauth/http4s/MauthPublicKeyProvider.scala
```diff
@@ -34,12 +35,15 @@ class MauthPublicKeyProvider[F[_]: Concurrent: Logger](configuration: Authentica
...
-      .flatMap(req => client.run(req).use(retrievePublicKey))
-  }
+    Async[F].defer {
+      val uri = new URI(configuration.getBaseUrl + getRequestUrlPath(appUUID))
+      val signedRequest = signer.signRequest(UnsignedRequest.noBody("GET", uri, headers = Map.empty))
+      signedRequest
+        .toHttp4sRequest[F]
+        .flatMap(req => client.run(req).use(retrievePublicKey))
+    }.memoize
+  }.flatten
```
jatcwang, line 45: If an error happened we should remove the IO from the cache.

tmilner: Yup, fair enough. Done that now.

modules/mauth-authenticator-http4s/src/main/scala/com/mdsol/mauth/http4s/MauthPublicKeyProvider.scala
```diff
@@ -23,23 +22,39 @@ import com.mdsol.mauth.http4s.MauthPublicKeyProvider.SecurityToken
...
     * @return { @link PublicKey} registered in MAuth for the application with given appUUID.
     */
-  override def getPublicKey(appUUID: UUID): F[Option[PublicKey]] = memoizeF(Some(configuration.getTimeToLive.seconds)) {
-    val uri = new URI(configuration.getBaseUrl + getRequestUrlPath(appUUID))
-    val signedRequest = signer.signRequest(UnsignedRequest.noBody("GET", uri, headers = Map.empty))
-    signedRequest
-      .toHttp4sRequest[F]
-      .flatMap(req => client.run(req).use(retrievePublicKey))
-  }
+  override def getPublicKey(appUUID: UUID): F[Option[PublicKey]] = cache
```
fserra-mdsol, line 36: How I would love that this was an actual `@newtype case class AppUUID(uuid: UUID)` instance, instead of a random `UUID` :-(

fserra-mdsol (approved):

jatcwang (approved):

PR #160, opened 2024-07-08
NOJIRA fix body consumption in http4s

fserra-mdsol:
As reported on [this http4s issue](https://github.com/http4s/http4s/issues/7443), in some conditions, if the request's EntityBody is consumed more than once by the caller, a `BodyAlreadyConsumedError` is raised. In our MAuthMiddleware, this can happen when performing the mAuth authentication on large or slow requests, whereby the request's Stream might be chunked, requiring multiple consumptions and causing intermittent failures.

This PR aims at solving this bug

### DIFF
--- a/modules/mauth-authenticator-http4s/src/main/scala/com/mdsol/mauth/http4s/MAuthMiddleware.scala
+++ b/modules/mauth-authenticator-http4s/src/main/scala/com/mdsol/mauth/http4s/MAuthMiddleware.scala
@@ -78,32 +78,32 @@ object MAuthMiddleware {
         else
           extractAll(V2) orElse extractAll(V1)
 
-      fk(request.as[Array[Byte]].flatMap { byteArray =>
-        authHeaderTimeHeader.flatMap { authCtx: MAuthContext =>
-          val mAuthRequest: MAuthRequest = new MAuthRequest(
-            authCtx.authHeader,
-            byteArray,
-            request.method.name,
-            authCtx.timeHeader.toString,
-            request.uri.path.renderString,
-            request.uri.query.renderString
-          )
-
-          // this mimics MAuthDirectives in the akka package - really needed?
-          val req = if (!authenticator.isV2OnlyAuthenticate) {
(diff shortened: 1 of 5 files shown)


amilne-mdsol (approved):

tmilner (approved):

modules/mauth-authenticator-http4s/src/main/scala/com/mdsol/mauth/http4s/MAuthMiddleware.scala
```diff
@@ -78,32 +78,32 @@ object MAuthMiddleware {
...
-          // this mimics MAuthDirectives in the akka package - really needed?
-          val req = if (!authenticator.isV2OnlyAuthenticate) {
-            mAuthRequest.setXmwsSignature(getHeaderValOrEmpty(V1.authHeaderName)) // dreadful mutating type
-            mAuthRequest.setXmwsTime(getHeaderValOrEmpty(V1.timeHeaderName))
-            mAuthRequest
-          } else mAuthRequest
-
-          authenticator.authenticate(req)(requestValidationTimeout).map(res => (res, authCtx))
+      fk(for {
+        strictBody <- request.toStrict(none)
```
tmilner, line 82: Actually, toStrict enforces a request size limit doesn't it? Could pass that in as an option (potentially with a default to keep it backwards compatible?)

fserra-mdsol: Nope, it doesn't enforce any limit - it would be recommended to enforce one, because `toStrict` only consumes the whole `EntityBody` and replaces any multipart header with a content-length (as a testament to its function), but there is no fixed size limit. In practice, this means that your application might blow up if the request is too large to be held in memory while it gets deserialised. I decided to not put any 'feature flag' on this, because I gauged that for our use cases there is no such risk - but if you know that it's not the case otherwise, I can add it.

PR #169, opened 2025-02-21
tech/MCC-1326235/UpdateDependenciesP2

jdepassos-mdsol:
dependencies update, sbt, scala, plugins

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -6,38 +6,38 @@ object Dependencies extends DependencyUtils {
     // Careful with upgrading akka due to license change https://www.lightbend.com/blog/why-we-are-changing-the-license-for-akka
     val akka: String = "2.6.20" // Do not update beyond 2.6.* due to license changes
     val akkaHttp: String = "10.2.10" // Do not update beyond 10.2.* due to license changes
-    val logback = "1.5.6"
-    val sttp = "3.10.2"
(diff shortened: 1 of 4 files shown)


project/Dependencies.scala
```diff
@@ -6,38 +6,38 @@ object Dependencies extends DependencyUtils {
...
+    val sttp = "3.10.3"
+    val http4s = "0.23.30"
+    val enumeratum = "1.7.5"
     val log4cats = "2.7.0"
-    val circe = "0.14.9"
-    val circeGenericExtras = "0.14.3"
-    val http4s022 = "0.22.15"
+    val circe = "0.14.10"
+    val circeGenericExtras = "0.14.4"
+    val http4s022 = "0.23.30"
```
tmilner, line 16: This says its http4s022? Do we have another for 023? Unsure if there is any need for continued 22 support, just checking

tmilner: Ah yeah, its seems like its supports both so this should be latest 0.22 version?

jdepassos-mdsol: 023 uses val http4s ...

mayman: further down there is Dsl22 & Dsl23 which is where the 022 package is used

mayman (approved):

